PR #20, opened 2022-02-02
Support allocation changes and networking quota

### LINKED ISSUE #1: [ColdFront] Implement quotas for OpenStack
Creating an issue so that this work can be tracked from CCI-MOC projects at https://github.com/orgs/CCI-MOC/projects

knikolla:
Closes #1 

- Persist current quota as allocation attribute on activation
- Also adds exit on error for run_functional_tests.sh
- Also adds a conditional to installing docker-compose for when
  running on vagrant.
- Added support for floating ips
- Refactored some of the common code for setting quotas
- Refactored the testing code for quotas to be more easy to read
- Renamed and updated utils.add_allocation_attribute to
  utils.get_allocation_attribute and now it supports updating
  existing allocation attributes
- Added a receiver for allocation_change_approval to
  activate_allocation.

### DIFF
--- a/src/coldfront_plugin_openstack/tasks.py
+++ b/src/coldfront_plugin_openstack/tasks.py
@@ -18,6 +18,28 @@ from coldfront_plugin_openstack import (attributes,
 
 logger = logging.getLogger(__name__)
 
+# Map the attribute name in ColdFront, to the client of the respective
+# service, the version of the API, and the key in the payload.
+QUOTA_KEY_MAPPING = {
+    'compute': {
+        'class': novaclient.Client,
+        'version': 2,
+        'keys': {
+            attributes.QUOTA_INSTANCES: 'instances',
+            attributes.QUOTA_VCPU: 'cores',
+            attributes.QUOTA_RAM: 'ram',
+        },
+    },
+    'volume': {
+        'class': cinderclient.Client,
+        'version': 3,
+        'keys': {
+            attributes.QUOTA_VOLUMES: 'volumes',
+            attributes.QUOTA_VOLUMES_GB: 'gigabytes',
+        }
+    }
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


src/coldfront_plugin_openstack/tasks.py
```diff
@@ -226,27 +251,31 @@ def create_default_network(resource, project_id):
 
 
 def activate_allocation(allocation_pk):
-    def set_nova_quota():
-        compute = novaclient.Client(NOVA_VERSION, session=get_session_for_resource(resource))
+    def set_quota():
```
jtriley, line 254: Just curious - why the nested function here?

jtriley (commented):
Looks good - couple of minor comments when you get a moment. Thanks!

test-requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/ubccr/coldfront@7dc4391c0cb54ec85ff82a0199ff0461454d5fb0#egg=coldfront
+git+https://github.com/ubccr/coldfront@2e60db247fd9a9b0d1299f1e98f1e4ef5f5c259b#egg=coldfront
```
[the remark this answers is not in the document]
knikolla: `coldfront-nerc` already has this commit version.

https://github.com/nerc-project/coldfront-nerc/blob/26877e499576a92757716236f190407a203fab6d/requirements.txt#L1

src/coldfront_plugin_openstack/tasks.py
knikolla: The nested function is accessing the attributes `allocation`, `resource`, and `openstack_project` from the outer scope. Reduced the complexity of passing attributes around the moment I was writing it, since `activate_allocation` is the only function calling `set_quotas`.

PR #22, opened 2022-02-02
Create default network when provisioning project

### LINKED ISSUE #21: [coldfront] Create default network for provisioned projects
As a user, I don't want to have to worry about the complexities of creating my first network when starting a project. I want to have one set up for me.

knikolla:
- Added attribute 'OpenStack Public Network ID'
- Added attribute 'OpenStack Default Network CIDR'
- Default network DNS set to Google DNS
- Added simple test case of network created

Closes #21

### DIFF
--- a/src/coldfront_plugin_openstack/tasks.py
+++ b/src/coldfront_plugin_openstack/tasks.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import secrets
 import urllib.parse
@@ -15,6 +16,7 @@ from novaclient import client as novaclient
 from coldfront_plugin_openstack import (attributes,
                                         utils)
 
+logger = logging.getLogger(__name__)
 
 NOVA_VERSION = '2'
 NOVA_KEY_MAPPING = {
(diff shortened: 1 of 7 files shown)


knikolla:
@larsks would appreciate a review from you on this one.

src/coldfront_plugin_openstack/tasks.py
```diff
@@ -137,6 +139,92 @@ def assign_role_on_user(resource, username, project_id):
...
+            }
+        }
+        network = neutron.create_network(body=default_network_payload)
+        logger.info(f'Default network with ID {network["network"]["id"]} '
+                    f'created for project {project_id}.')
+
+    # Get or create default subnet
+    subnets = neutron.list_subnets(project_id=project_id,
+                                   name='default_subnet')
+    if subnets := subnets['subnets']:
```
larsks, line 168: Do you need to worry about the situation in which a `default_subnet` exists but it is associated with something other than `default_network`?

larsks (approved):
Looks good to me, although I left one question about the degree of error checking which is appropriate for this code.

src/coldfront_plugin_openstack/tasks.py
knikolla: Good question. Given that this code will only run on new project creations, I don't think we do need to worry about that. This error checking is mostly to protect against half-executions of itself and ensure a certain degree of idempotency.

src/coldfront_plugin_openstack/tests/base.py
```diff
@@ -103,6 +105,12 @@ def new_resource(name=None, auth_url=None) -> Resource:
             resource=openstack,
             value=1
         )
+        ResourceAttribute.objects.get_or_create(
+            resource_attribute_type=ResourceAttributeType.objects.get(
+                name=attributes.RESOURCE_DEFAULT_PUBLIC_NETWORK),
+            resource=openstack,
+            value=os.getenv('OPENSTACK_PUBLIC_NETWORK_ID')
```
jtriley, line 112: Should this be parsed similar to `OPENSTACK_DEVSTACK_APPLICATION_CREDENTIAL_ID`? For example if there are multiple openstack clusters...

jtriley:
Need up update the README with these new options.

PR #35, opened 2022-03-23
Set Swift Quotas

knikolla:
This PR adds a new attribute for quotas 'OpenStack Swift Quota in Gigabytes' and adds the logic to set the quota on Swift. Requires the `account-quotas` middleware in the swift proxy pipeline.

### DIFF
--- a/ci/devstack.sh
+++ b/ci/devstack.sh
@@ -3,7 +3,8 @@
 #
 set -xe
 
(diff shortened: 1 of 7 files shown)


jtriley:
> Requires the account-quotas middleware in the swift proxy pipeline.

Does this mean we need to update swift deployment parameters to enable that on the OpenStack side?

PR #52, opened 2022-06-08
Rename attributes in register_cloud_attributes command

### LINKED ISSUE #42: [ColdFront] Create management command to rename attributes
If an attribute changes name or characteristics between plugin versions, the new version is going to duplicate it. Create a management command to help with "migrations".

knikolla:
It is now possible to define a list of attributes that will be
automatically renamed when running the above management command.

Additionally, this creates a new github action to run the
newly necessary unit tests.

Closes #42

### DIFF
--- a/.github/workflows/test-py39-unit.yaml
+++ b/.github/workflows/test-py39-unit.yaml
@@ -1,0 +1,27 @@
+name: test-py39-unit
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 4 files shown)


larsks (approved):
Looks good.

Out of curiosity, does the `handle` method run periodically, or is this somehow event triggered?

larsks:
@knikolla Although I have approved this pr, github stills shows a pending review request. In any case, I have no further comments :)

knikolla:
> Looks good.
> 
> Out of curiosity, does the `handle` method run periodically, or is this somehow event triggered?

The `handle` method is what gets called by Django when you run the command the module is named as. So in this case running `coldfront register_cloud_attributes` from the command line triggers execution.

jtriley (approved):

PR #63, opened 2022-08-03
Allow Resource Allocators to define a project name character limit

### LINKED ISSUE #53: To restrict project title max length to 64 characters
- Right now user can input more than 64 characters as project title.
- Need to restrict it so user will not allowed to input lengthy title.
- validation on client and server side.

knikolla:
Closes #53

### DIFF
--- a/src/coldfront_plugin_cloud/tasks.py
+++ b/src/coldfront_plugin_cloud/tasks.py
@@ -60,8 +60,10 @@ def find_allocator(allocation) -> base.ResourceAllocator:
         return allocator_class(resource, allocation)
 
 
-def get_unique_project_name(project_name):
-    return f'{project_name}-f{secrets.token_hex(3)}'
+def get_unique_project_name(project_name, max_length=None):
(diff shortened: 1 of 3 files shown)


larsks (approved):

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -60,8 +60,10 @@ def find_allocator(allocation) -> base.ResourceAllocator:
         return allocator_class(resource, allocation)
 
 
-def get_unique_project_name(project_name):
-    return f'{project_name}-f{secrets.token_hex(3)}'
+def get_unique_project_name(project_name, max_length=None):
+    # The random hex at the end of the project name is 6 chars
+    max_without_suffix = max_length - 6 if max_length else None
```
jtriley, line 65: Should this be `7` instead to include the hyphen?

knikolla: ah, good catch!

jtriley (approved):

PR #64, opened 2022-08-03
Populate missing attributes on resource allocations

### LINKED ISSUE #41: [ColdFront] Create management cmd to populate missing attributes on resource allocations
When a new attribute is added to a resource allocation, previously activated resource allocations are going to be missing the attribute.

knikolla:
The attribute will be set based on the current quota value.

Closes #41

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -2,6 +2,7 @@ import logging
 
 from coldfront_plugin_cloud import attributes
 from coldfront_plugin_cloud import openstack
+from coldfront_plugin_cloud import utils
 
 from django.core.management.base import BaseCommand
 from coldfront.core.resource.models import (Resource,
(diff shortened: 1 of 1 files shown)


naved001 (approved):

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -58,22 +59,29 @@ def handle(self, *args, **options):
...
+                    if expected_value is None and current_value:
+                        msg = (f'Attribute "{attr}" expected on allocation {allocation_str} but not set.'
+                               f' Current quota is {current_value}.')
+                        if options['apply']:
+                            utils.set_attribute_on_allocation(
+                                allocation, attr, current_value
+                            )
+                            msg = f'{msg} Attribute set to match current quota.'
+                        logger.warning(msg)
+                    elif not current_value == expected_value:
```
larsks, line 80: This may be a dumb question: by `expected_value` do we actually mean `requested_value`?

knikolla: Yes. Both terminologies are correct. I chose *expected* because from the perspective of ColdFront, that requested quota should have already been applied, but for some reason the system is in an inconsistent/unexpected state.

jtriley (approved):

PR #65, opened 2022-08-03
Rename attributes to remove OpenStack prefix

### LINKED ISSUE #59: Rename project_id and project_name attributes to remove the OpenStack prefix
https://github.com/nerc-project/coldfront-plugin-openstack/pull/52 introduced attribute name migrations.

The project name and ID attributes contain an OpenStack prefix, however they are now not openstack specific and also apply to OpenShift (and other possible services supported by the plugin).

Use the attribute name migration feature to remove away the prefix.

knikolla:
The following Resource attributes are migrated
- OpenStack Auth URL -> Identity Endpoint URL
- OpenStack Role for User in Project -> Role for User in Project

The following Allocation attributes are migrated
- OpenStack Project Name -> Allocated Project Name
- OpenStack Project ID -> Allocated Project ID

The choice of allocated is to signal that the allocation has
been completed on the service.

Additionaly, a test has been added to verify that the migration
of resource attributes preserves the value of the attribute.

Closes #59

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -1,8 +1,9 @@
-RESOURCE_AUTH_URL = 'OpenStack Auth URL'  # TODO: remove OpenStack prefix
+RESOURCE_AUTH_URL = 'Identity Endpoint URL'
+RESOURCE_ROLE = 'Role for User in Project'
+
 RESOURCE_FEDERATION_PROTOCOL = 'OpenStack Federation Protocol'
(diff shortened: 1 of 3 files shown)


src/coldfront_plugin_cloud/attributes.py
```diff
@@ -1,4 +1,4 @@
-RESOURCE_AUTH_URL = 'OpenStack Auth URL'  # TODO: remove OpenStack prefix
```
knikolla, line 1: Need to remove the comment. Will do after CI runs.

jtriley (approved):

PR #66, opened 2022-08-12
Email on task error

### LINKED ISSUE #58: When a Django-Q task fails, an email should be sent notifying the admin of the failure

knikolla:
Send an email when tasks encounter an exception.

Closes #58

### DIFF
--- a/src/coldfront_plugin_cloud/signals.py
+++ b/src/coldfront_plugin_cloud/signals.py
@@ -1,4 +1,5 @@
 import os
+import traceback
 
 from django.dispatch import receiver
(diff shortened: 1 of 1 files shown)


jtriley:
It might be worth looking into using a django-q error_reporter for this functionality, e.g. 

https://django-q.readthedocs.io/en/latest/errors.html#errors
https://django-q.readthedocs.io/en/latest/configure.html#error-reporter
https://github.com/danielwelch/django-q-rollbar

knikolla:
@jtriley that seems like a much better approach and pretty simple to implement. Thanks for looking it up.

PR #67, opened 2022-09-01
fix swift quotas

jtriley:
- Rework how we "switch projects" to avoid templating issues between various endpoint URL implementations (e.g. devstack using `$` vs RHOSP using `%` for template variables). This issue has resulted in 403 forbidden errors connecting to swift on RHOSP.
- Handle case where swift `head_account` is missing the quota key


jtriley:
As expected this should fail and did with the current code:
```
======================================================================
FAIL: test_new_allocation (coldfront_plugin_cloud.tests.functional.openstack.test_allocation.TestAllocation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/coldfront-plugin-cloud/coldfront-plugin-cloud/src/coldfront_plugin_cloud/tests/functional/openstack/test_allocation.py", line 103, in test_new_allocation
    self.assertEqual(actual_swift_quota.get(k), v)
AssertionError: '1000000000' != 1000000000
```
Adding the fix now to see if it passes...

knikolla (approved):

Issue #68, opened 2022-09-14
Create management command to update EULA from URL


knikolla:
As an administrator, I want the terms of service that my users have to accept to be up to date and fetched from a central location where they are stored in plaintext format.

Create a new attribute on the resource called `eula_url` and create a new management command that fetches the contents of that location and updates the `eula` attribute on the resource.

naved001:
Hi @knikolla here's a draft of what I have:

https://github.com/naved001/coldfront-plugin-cloud/blob/add-update-eula-command/src/coldfront_plugin_cloud/management/commands/update_eula.py

Though at this point it needs to know the resource name and resource type it's updating the eula for (instead of going over all resources).

knikolla:
Hey Naved, look pretty good. Some minor feedback

1. Is the resource_type argument really required? I think just name should be enough. 
2. Please check the status of the http request and return an error if it doesn't fetch correctly.
3. There's a `get_or_create` method that would merge your try/catch block https://github.com/nerc-project/coldfront-plugin-cloud/blob/6978fc68f3356ff74400419e06883cd9a64801d5/src/coldfront_plugin_cloud/management/commands/add_openshift_resource.py#L40

naved001:
> Is the resource_type argument really required? I think just name should be enough.

I guess not if I can get the resource object just by name. 

> There's a get_or_create method that would merge your try/catch block 

I tried that and the similar `update_or_create` method, but on a subsequent run I was getting SQL Integrity error and I couldn't actually update it. I will post the full error message later.

naved001:
Here's the full error when I use get_or_create (or update_or_create). The error only shows up when it actually has to update the value i.e. if I re-run the command with the same url then nothing happens.

```
Traceback (most recent call last):
  File "/tmp/coldfront_venv/lib/python3.8/site-packages/django/db/models/query.py", line 581, in get_or_create
    return self.get(**kwargs), False
  File "/tmp/coldfront_venv/lib/python3.8/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
coldfront.core.resource.models.DoesNotExist: ResourceAttribute matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/coldfront_venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
[...]

PR #70, opened 2022-11-07
Add update eula command

naved001:
Currently this takes the resource name as an argument. Let me know if we need to update all Resources wholesale instead.

closes https://github.com/nerc-project/coldfront-plugin-cloud/issues/68

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/update_eula.py
+++ b/src/coldfront_plugin_cloud/management/commands/update_eula.py
@@ -1,0 +1,52 @@
+import requests
+
+from django.core.management.base import BaseCommand
+
+from coldfront.core.resource.models import (
+    Resource,
+    ResourceAttribute,
+    ResourceAttributeType,
+    ResourceType,
+)
+
+from coldfront_plugin_cloud import attributes
+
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/management/commands/update_eula.py
```diff
@@ -0,0 +1,54 @@
...
+from coldfront.core.resource.models import (
+    Resource,
+    ResourceAttribute,
+    ResourceAttributeType,
+    ResourceType,
+)
+
+from coldfront_plugin_cloud import attributes
+
+EULA = "eula"
```
knikolla, line 14: A more descriptive name for this variable would be `TARGET_ATTRIBUTE`. Took me a second to figure out what the use of this was in the current state.

knikolla (approved):
Looks good, minor comment inline :)

src/coldfront_plugin_cloud/management/commands/update_eula.py
naved001: Done.

jtriley (approved):

PR #71, opened 2022-11-17
fix preauth url config in swift client

jtriley:
Turns out, replacing the url attribute is not the same as passing preauthurl. Removed the swiftclient_change_project function in favor of simply constructing the initial swift client with the proper preauthurl kwarg if a project_id is specified.


knikolla (approved):
approved, pending CI.

PR #76, opened 2022-12-12
Extend validation command to validate openshift quotas

naved001:
this does the thing, but could look nicer and maybe use some refactoring. Looking for feedback.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -1,10 +1,12 @@
 import logging
+import re
 
 from coldfront_plugin_cloud import attributes
 from coldfront_plugin_cloud import openstack
+from coldfront_plugin_cloud import openshift
 from coldfront_plugin_cloud import utils
 
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 from coldfront.core.resource.models import (Resource,
                                             ResourceType)
 from coldfront.core.allocation.models import (Allocation,
@@ -23,16 +25,18 @@ class Command(BaseCommand):
                             help='Apply expected state if validation fails.')
 
     def handle(self, *args, **options):
-        supported_resources = Resource.objects.filter(
+
+        # Openstack Resources first
+        openstack_resources = Resource.objects.filter(
             resource_type=ResourceType.objects.get(
                 name='OpenStack'
             )
         )
-        allocations = Allocation.objects.filter(
-            resources__in=supported_resources,
+        openstack_allocations = Allocation.objects.filter(
+            resources__in=openstack_resources,
             status=AllocationStatusChoice.objects.get(name='Active')
         )
-        for allocation in allocations:
+        for allocation in openstack_allocations:
             allocation_str = f'{allocation.pk} of project "{allocation.project.title}"'
             msg = f'Starting resource validation for allocation {allocation_str}.'
             logger.debug(msg)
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -69,22 +74,134 @@ def handle(self, *args, **options):
...
+        )
+        openshift_allocations = Allocation.objects.filter(
+            resources__in=openshift_resources,
+            status=AllocationStatusChoice.objects.get(name="Active"),
+        )
+
+        for allocation in openshift_allocations:
+            allocation_str = f'{allocation.pk} of project "{allocation.project.title}"'
+            msg = f"Starting resource validation for allocation {allocation_str}."
+            logger.debug(msg)
```
larsks, line 116: I don't think I would bother with a throwaway variable here; just write:

```
logger.debug(f"...the message...")
```

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -69,22 +74,134 @@ def handle(self, *args, **options):
...
+                    map_to_Gi = {
+                        "Ki": 2**-20,
+                        "Mi": 2**-10,
+                        "Gi": 2**0,
+                        "Ti": 2**10,
+                    }
+
+                    try:
+                        if current_value and "RAM" in attr:
+                            result = re.search(PATTERN, str(current_value))
```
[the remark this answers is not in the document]
larsks: Also, since you're making the same `re.search` call in both this case and the following one, why not move `result = re.search(PATTERN, str(current_value))` line out of the `if` statement?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -69,22 +74,134 @@ def handle(self, *args, **options):
...
+
+            for attr in attributes.ALLOCATION_QUOTA_ATTRIBUTES:
+                if "OpenShift" in attr:
+                    key_with_lambda = openshift.QUOTA_KEY_MAPPING.get(attr, None)
+
+                    # This gives me just the plain key
+                    key = list(key_with_lambda(1).keys())[0]
+
+                    expected_value = allocation.get_attribute(attr)
+                    current_value = quota.get(key, None)
```
larsks, line 148: What data type is returned here? You're treating `current_value` as a string value when matching it against a regular expression, but later on you're setting it to an `int` value.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -69,22 +74,134 @@ def handle(self, *args, **options):
...
+                                result.groups()[1]
+                            )
+                        elif current_value:
+                            current_value = int(current_value)
+                    except AttributeError:
+                        CommandError(
+                            f"For attribute '{attr}' could not not parse current_value = {current_value}"
+                        )
+
+                    if expected_value is None and current_value:
```
larsks, line 184: Can `current_value` ever legitimately be `0`?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -69,22 +74,134 @@ def handle(self, *args, **options):
...
+                            )
+                            msg = f"{msg} Attribute set to match current quota."
+                        logger.warning(msg)
+                    elif not current_value == expected_value:
+                        failed_validation = True
+                        msg = (
+                            f"Value for quota for {attr} = {current_value} does not match expected"
+                            f" value of {expected_value} on allocation {allocation_str}"
+                        )
+                        logger.warning(msg)
```
larsks, line 201: Why don't we check for `options["apply"]` in this case (and reset the quota if necessary)?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -69,22 +74,134 @@ def handle(self, *args, **options):
...
                     current_value = quota.get(key, None)
                     if expected_value is None and current_value:
-                        msg = (f'Attribute "{attr}" expected on allocation {allocation_str} but not set.'
-                               f' Current quota is {current_value}.')
-                        if options['apply']:
+                        msg = (
+                            f'Attribute "{attr}" expected on allocation {allocation_str} but not set.'
+                            f" Current quota is {current_value}."
+                        )
+                        if options["apply"]:
```
larsks, line 81: It looks like you've applied some reformatting here. It would be helpful if you either (a) omitted these changes, or (b) moved them into a separate commit (because it makes it hard to tell what changes here are operationally significant).

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: good point, I'll try setting it and see what happens!

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: `current_value` here is a string for all attributes e.g. it could be `"1000Mi"` for RAM, `"20Gi"` for Storage, or `"4"` for CPU. 

So for RAM and Storage quota, I use the regular expression to extract the value and unit, and ultimately get an `int` value of the right unit. I need all to be numbers because that's how coldfront stores these.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: oops, I'll fix this. Been trying out Visual Studio code with  Black formatting enabled, and I guess it formatted stuff when I saved :(

naved001:
Thank you for the review @larsks!

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
larsks: I think it makes sense to rigorously format everything with Black (and make that part of the CI checks). Just keep the formatting changes in separate commits.

naved001: okay, I took care of that.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: We set a boolean `failed_validation`in this block  to `True` and then in a following if statement checks for that boolean and `options["apply"]` to rest the quota. But I think we can just stick the check right here and get rid of setting the unnecessary boolean since that's not used elsewhere.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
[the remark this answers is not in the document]
naved001: > What if re.search() fails to find a match?

Then the next call to `result.groups()[0]` will trigger an AttributeError
```
AttributeError: 'NoneType' object has no attribute 'groups'
```
which is then re-raised as a CommandError.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: OpenShift did let me set the quota value to 0. So`current_value` before this block can be 0 (instead of None), specifically a string `"0"`.

For RAM and Storage, an AttributeError earlier will stop the execution. For CPU, it'll get cast to an `int` and `bool(int("0"))` will return False and it won't go into this block at all.

So I guess I need to handle the case of values being 0 better. Also, `m` is a valid unit for memory apparently which I am not handling

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: I have now updated this to do the check and reset the quota if necessary.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: if `current_value` is `"0"` and `expected_value` is `None`, then we will end up setting the quota to "0" in coldfront, since we don't really know what the actual value should be anyway. It'll at least make it match.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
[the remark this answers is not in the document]
naved001: I have since updated the code and I am raising a CommandError when re.search() fails to find a match.

larsks (approved):

knikolla (changes requested):
Good work! The change looks good to me, but I have a few comments.

First, I want to see it tested before merging. Example of how this is done with OpenStack is here https://github.com/nerc-project/coldfront-plugin-cloud/blob/b5dc67c5e235577667cdf9966b8663e0cf989978/src/coldfront_plugin_cloud/tests/functional/openstack/test_allocation.py#L186-L198

Second, this project doesn't use `black` yet, and I'm not happy with the automatic introduction of unnecessary code restructuring in parts of the file not touched by the scope of this PR.

naved001:
@knikolla I removed the formatting changes and added a test similar to that of openstack. Let me know if you want more comprehensive tests.

knikolla:
Thanks! I'll re-review this evening.

knikolla (approved):
Looks good to me. Thanks @naved001!

PR #84, opened 2023-02-01
Update pip package in CI setup

knikolla:
An update to `setuptools` broke `pip`. This updates `pip` to pick up the fix, and in turn fix our CI.


jtriley (approved):

rob-baron (commented):
looks goog.

PR #88, opened 2023-02-16
Include Coldfront project name when generating project display name

naved001:
This commit moves `get_unique_project_name` to `utils.py` and then use that when generating the name. It also adds a check to the method to validate the project name.

### DIFF
--- a/src/coldfront_plugin_cloud/openshift.py
+++ b/src/coldfront_plugin_cloud/openshift.py
@@ -5,7 +5,6 @@ import os
 import requests
 from requests.auth import HTTPBasicAuth
 import time
-import uuid
 from simplejson.errors import JSONDecodeError
 
 from coldfront_plugin_cloud import attributes, base, utils
@@ -33,6 +32,8 @@ class Conflict(ApiException):
 class OpenShiftResourceAllocator(base.ResourceAllocator):
 
     resource_type = 'openshift'
+
+    project_name_max_length = 63
 
     @functools.cached_property
     def session(self):
@@ -68,10 +69,14 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
         else:
             raise ApiException(f"{response.status_code}: {response.text}")
 
-    def create_project(self, project_name):
-        project_id = uuid.uuid4().hex
+    def create_project(self, suggested_project_name):
+        sanitized_project_name = utils.get_sanitized_project_name(suggested_project_name)
+        project_id = utils.get_unique_project_name(
+            sanitized_project_name,
+            max_length=self.project_name_max_length)
+        project_name = project_id
         self._create_project(project_name, project_id)
-        return project_id
+        return self.Project(project_name, project_id)
 
     def set_quota(self, project_id):
         url = f"{self.auth_url}/projects/{project_id}/quota"
--- a/src/coldfront_plugin_cloud/openstack.py
+++ b/src/coldfront_plugin_cloud/openstack.py
@@ -119,13 +119,17 @@ class OpenStackResourceAllocator(base.ResourceAllocator):
             preauthurl=preauth_url,
         )
 
-    def create_project(self, project_name) -> str:
+    def create_project(self, suggested_project_name) -> base.ResourceAllocator.Project:
+        project_name = utils.get_unique_project_name(
+            suggested_project_name,
+            max_length=self.project_name_max_length)
+
         openstack_project = self.identity.projects.create(
             name=project_name,
             domain=self.resource.get_attribute(attributes.RESOURCE_PROJECT_DOMAIN),
             enabled=True,
         )
-        return openstack_project.id
+        return self.Project(project_name, openstack_project.id)
 
     def reactivate_project(self, project_id):
         openstack_project = self.identity.projects.get(project_id)
--- a/src/coldfront_plugin_cloud/tasks.py
+++ b/src/coldfront_plugin_cloud/tasks.py
@@ -60,12 +60,6 @@ def find_allocator(allocation) -> base.ResourceAllocator:
         return allocator_class(resource, allocation)
 
 
-def get_unique_project_name(project_name, max_length=None):
-    # The random hex at the end of the project name is 6 chars, 1 hyphen
-    max_without_suffix = max_length - 7 if max_length else None
-    return f'{project_name[:max_without_suffix]}-f{secrets.token_hex(3)}'
-
-
 def activate_allocation(allocation_pk):
     def set_quota_attributes():
         if allocation.quantity < 1:
@@ -88,11 +82,10 @@ def activate_allocation(allocation_pk):
         if project_id := allocation.get_attribute(attributes.ALLOCATION_PROJECT_ID):
             allocator.reactivate_project(project_id)
         else:
-            project_name = get_unique_project_name(
-                allocation.project.title,
-                max_length=allocator.project_name_max_length
-            )
-            project_id = allocator.create_project(project_name)
+            project = allocator.create_project(allocation.project.title)
+
+            project_id = project.id
+            project_name = project.name
 
             utils.set_attribute_on_allocation(allocation,
                                               attributes.ALLOCATION_PROJECT_NAME,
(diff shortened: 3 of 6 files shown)


knikolla (changes requested):
This isn't going to work because we don't have control over what is inputted as a Project Name in ColdFront. They can include whitespace, and special characters.

For example here is an actual project name that I picked from the current ColdFront system. `TEST - Software & Application Innovation Lab (SAIL) - TEST Projects`

```
>>> get_unique_project_name("TEST - Software & Application Innovation Lab (SAIL) - TEST Projects")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in get_unique_project_name
AssertionError: not a valid project name
```

Also OpenStack doesn't have those limitations (except length), therefore it is desirable to keep the complexity of the name if possible. So it should preserve the current behavior for OpenStack project names. 

You need to provide a method to convert the arbitrary project name provided as an example above to a namespace name that is valid for OpenShift, and use that for the project ID and project Name for OpenShift only.

naved001:
@knikolla I have updated this PR.

```
>>> get_sanitized_project_name("TEST - Software & Application Innovation Lab (SAIL) - TEST Projects")
'test---software---application-innovation-lab--sail----test-projects'
```
Note that this is longer than 64 characters, the other method takes care of the length and makes it unique.

> and use that for the project ID and project Name for OpenShift only.

The project display-name does not have to conform to a spec, only the actual name (project_id).  But I am using the sanitized name for display-name anyway. 

```
naved@computer ~ % oc get ns delete-me-because-this-is-a-test -o yaml |oc neat
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/display-name: Display N@me has bunch of test characters and may be
      a bit too long, does this have a character limitation, we'll see soon
```

src/coldfront_plugin_cloud/utils.py
```diff
@@ -22,3 +24,17 @@ def set_attribute_on_allocation(allocation, attribute_type, attribute_value):
...
             value=attribute_value,
         )
+
+def get_unique_project_name(project_name, max_length=None):
+    # The random hex at the end of the project name is 6 chars, 1 hyphen
+    max_without_suffix = max_length - 7 if max_length else None
+    return f'{project_name[:max_without_suffix]}-f{secrets.token_hex(3)}'
+
+def get_sanitized_project_name(project_name):
+    # convert to lowercase
```
knikolla, line 34: This comment is redundant since the following line provides the same exact information. 

Suggestion: It would be more helpful here to provide the motivation and limitations that this method is trying to work around, rather than comment individual lines (except regex, always comment regex.)

For example:

> Returns a sanitized project name that only contains lowercase alphanumeric characters and dashes (not leading or trailing.)

knikolla:
> @knikolla I have updated this PR.
> 
> ```
> >>> get_sanitized_project_name("TEST - Software & Application Innovation Lab (SAIL) - TEST Projects")
> 'test---software---application-innovation-lab--sail----test-projects'
> ```
> 

That looks almost as good as we're ever going to get. Though it might make sense to strip consecutive dashes to save on characters.

src/coldfront_plugin_cloud/utils.py
naved001: makes sense, I have updated it.

naved001:
> Though it might make sense to strip consecutive dashes to save on characters.

@knikolla good point, updated to do that.

```
>>> get_sanitized_project_name("TEST - Software & Application Innovation Lab (SAIL) - TEST Projects")
'test-software-application-innovation-lab-sail-test-projects'
```

src/coldfront_plugin_cloud/utils.py
joachimweyl: @knikolla is this resolved?

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -88,7 +82,7 @@ def set_quota_attributes():
         if project_id := allocation.get_attribute(attributes.ALLOCATION_PROJECT_ID):
             allocator.reactivate_project(project_id)
         else:
-            project_name = get_unique_project_name(
+            project_name = utils.get_unique_project_name(
```
knikolla, line 85: There's a few things wrong here.

You're calling `get_unique_project_name`, which is trimming it down and adding a random hex at the end, however when you call `create_project` it calls the above method again. Therefore you're ending up with double the random hex at the end for project id, and the display name will not match the project id.

Also, the project name defined here is going to be added to the attribute `Allocated Project Name` and that is going to be different from the name that was assigned as the display name, with the latter being sanitized and this one not.

knikolla (changes requested):

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -68,10 +67,11 @@ def check_response(response: requests.Response):
         else:
             raise ApiException(f"{response.status_code}: {response.text}")
 
-    def create_project(self, project_name):
-        project_id = uuid.uuid4().hex
+    def create_project(self, suggested_project_name):
+        project_name = utils.get_sanitized_project_name(suggested_project_name)
+        project_id = utils.get_unique_project_name(project_name)
```
knikolla, line 72: You still need to provide a max length here. The base class provides a `project_name_max_length` which we can set in the OpenShift allocator and pass here.

knikolla (changes requested):

src/coldfront_plugin_cloud/openshift.py
naved001: oops, thanks for pointing that out! Fixed it,

naved001: actually wait, let me do it the right way

naved001: done.

knikolla: It looks okay now after the revision, but I'd use the same output from `get_unique_project_name` here for both `project_name` and `project_id`, it will not contain the randomness at the end and it makes sense for consistency.

knikolla:
Made a last minor comment in the thread, outside of that looks great! Thanks!

src/coldfront_plugin_cloud/openshift.py
naved001: > it will not contain the randomness at the end

You mean it will contain the randomness for project_name?

knikolla: Ah, I poorly phrased it. Yes. I meant that in the current state `project_name` does not contain the randomness and is different from the namespace name.

So, just set project_name and the display name to whatever we got for project_id and return that.

src/coldfront_plugin_cloud/openstack.py
```diff
@@ -119,13 +119,17 @@ def object(self, project_id=None) -> swiftclient.Connection:
             preauthurl=preauth_url,
         )
 
-    def create_project(self, project_name) -> str:
+    def create_project(self, suggested_project_name) -> base.ResourceAllocator.Project:
```
naved001, line 122: @knikolla question: In a method in a derived class, does one really need to indicate the return type? Doesn't having that in the base class sufficient?

We also only did this for openstack ,and not for openshift. So I only updated this here.

src/coldfront_plugin_cloud/openshift.py
naved001: thanks, I have made the change.

src/coldfront_plugin_cloud/openstack.py
knikolla: Great question. Player around with it.

```python
import abc


class BaseClass(abc.ABC):
    @abc.abstractmethod
    def method(self) -> int:
        pass


class SubClass(BaseClass):
    def method(self):
        return 1

>>> typing.get_type_hints(SubClass.method)
{}
>>> typing.get_type_hints(BaseClass.method)
{'return': <class 'int'>}
```

So, for code that treats instances of `SubClass` as `BaseClass` the type hints are going to work, but for code that instantiates the subclass directly, they aren't.

knikolla (approved):

PR #92, opened 2023-03-27
Make it easier to run CI locally

larsks:
This is a series of changes to make it easier to run the CI tests in a local Linux environment. In particular:

- We move more configuration into environment variables
- We move distribution-specific package installs to a separate script
- We avoid conflicts with host services running on the same ports used by Microshift and the image registry
- We make it possible to run the tests without clobbering a local `~/.kube/config` file

### DIFF
--- a/ci/microshift.sh
+++ b/ci/microshift.sh
@@ -11,17 +11,24 @@ test_dir="$PWD/testdata"
 rm -rf "$test_dir"
 mkdir -p "$test_dir"
 
-echo '127.0.0.1  onboarding-onboarding.cluster.local' | sudo tee -a /etc/hosts
+sudo docker rm -f microshift registry
+
+registry_port=$(( RANDOM % 1000 + 10000 ))
 
 echo "::group::Start microshift container"
 sudo docker run -d --rm --name microshift --privileged \
-    --network host \
+    --hostname microshift \
     -v microshift-data:/var/lib \
+    -p "${registry_port}:5000" \
     quay.io/microshift/microshift-aio:latest
 echo "::endgroup::"
 
+microshift_addr=$(sudo docker inspect microshift -f '{{ .NetworkSettings.IPAddress }}')
+sudo sed -i '/onboarding-onboarding.cluster.local/d' /etc/hosts
+echo "$microshift_addr  onboarding-onboarding.cluster.local" | sudo tee -a /etc/hosts
+
 echo "::group::Start registry container"
-sudo docker run -d --name registry --network host registry:2
+sudo docker run -d --name registry --network container:microshift registry:2
 echo "::endgroup::"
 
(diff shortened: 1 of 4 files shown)


ci/setup-ubuntu.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+set -xe
+
+sudo apt-get update && sudo apt-get upgrade -y
+sudo apt-get install docker-compose python3-virtualenv
```
naved001, line 6: why does this command run in the CI without the `-y` flag?

Another thing I noticed is that the package `docker-compose` recommends `docker.io` so apt will install it if it's missing on the system unless one specifies the `-no-install-recommends` flag (I am not saying we should use that flag, just thought it was interesting).

naved001 (approved):

ci/setup-ubuntu.sh
larsks: It should have the `-y` flag.

Good spotting re: `docker-compose`. Since we're not using it, we should just remove it; previously, the `microshift.sh` script was explicitly installing `docker.io`, but GitHub runners already include Docker. I hadn't noticed the dependency.

ci/microshift.sh
```diff
@@ -3,45 +3,65 @@
...
 
-# https://github.com/nerc-project/coldfront-plugin-cloud/issues/50
-sleep 30
+echo "::group::Start registry container"
+sudo docker run -d --name registry --network container:microshift registry:2
+echo "::endgroup::"
 
-curl -O "https://mirror.openshift.com/pub/openshift-v4/$(uname -m)/clients/ocp/stable/openshift-client-linux.tar.gz"
-sudo tar -xf openshift-client-linux.tar.gz -C /usr/local/bin oc kubectl
+mkdir -p ~/.kube
```
knikolla, line 34: You should use the environment variable `KUBECONFIG` here to figure out where to create the directory.

ci/microshift.sh
```diff
@@ -3,45 +3,65 @@
...
 done
+echo "::endgroup::"
 
 # Install OpenShift Account Management
-git clone https://github.com/cci-moc/openshift-acct-mgt.git ~/openshift-acct-mgt
-cd ~/openshift-acct-mgt
-git checkout "$ACCT_MGT_VERSION"
-sudo docker build . -t "localhost:5000/cci-moc/openshift-acct-mgt:latest"
-sudo docker push "localhost:5000/cci-moc/openshift-acct-mgt:latest"
+git config --global advice.detachedHead false
```
knikolla, line 53: Since this PR seems to provide some guaranteed regarding being run locally on the developer's environment, I don't feel comfortable changing their global git config.

knikolla (changes requested):

knikolla (approved):

PR #93, opened 2023-03-31
Initial REST API to list Allocations

### LINKED ISSUE #89: Create HTTP endpoint to read current projects, PI, resource allocations, and users
To map the usage data in XdMod with who to bill for it, we need to provide a way for ColdFront to export the relationship between the following resources

- Projects
- Resource Allocations
- PI
- Users in Project

----------
Edited Original description
- Allow admin to export the current projects in excel and csv format from admin dashboard.
- To show graphs of all current projects on the Center Summary screen

knikolla:
This is roughly based on https://github.com/ubccr/coldfront/pull/425
and creates a simple API for listing active resource allocations at
/cloud-api/allocations and /cloud-api/allocations/<id>.

---

The first commit introduces the API using djangorestframework, and required switching to a
fork of ColdFront to allow extending the routes to include the routes introduced by the plugin.
https://github.com/knikolla/coldfront/commit/edf99cf3104ee96ad7d3d27dabb0b297921344ea

---

The second commit limits the Allocations REST API to admin users.

In the presence of the PLUGIN_OIDC flag, which signals that ColdFront
has been configured with OpenID Connect, this only allows
authentication via access tokens.

If the flag has not been detected, than it falls back to
Basic and Session authentications.

---

One thing of note, is that this is based on the current state of master and one of the changes there is [introducing a confirmation dialog](https://github.com/ubccr/coldfront/pull/460/files) to `coldfront initial_setup`, so I've had to update some of the testing code to take that into account. This will have no consequences on `coldfront-nerc` as we're not planning on running `initial_setup` or scripting it's execution there.

Closes #89

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-coldfront >= 1.0.4
+git+https://github.com/knikolla/coldfront@coldfront_plugin_cloud_urls#egg=coldfront
+djangorestframework
 python-cinderclient  # TODO: Set version for OpenStack Clients
 python-keystoneclient
 python-memcached==1.59
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 git+https://github.com/knikolla/coldfront@coldfront_plugin_cloud_urls#egg=coldfront
+mozilla-django-oidc
 djangorestframework
 python-cinderclient  # TODO: Set version for OpenStack Clients
 python-keystoneclient
--- a/src/coldfront_plugin_cloud/urls.py
+++ b/src/coldfront_plugin_cloud/urls.py
@@ -1,13 +1,25 @@
+import os
+
 from rest_framework import routers, viewsets
+from rest_framework.authentication import SessionAuthentication, BasicAuthentication
+from rest_framework.permissions import IsAdminUser
+from mozilla_django_oidc.contrib.drf import OIDCAuthentication
 
 from coldfront.core.allocation.models import Allocation
 
 from coldfront_plugin_cloud.api import serializers
 
+if os.getenv('PLUGIN_OIDC') == 'True':
+    AUTHENTICATION_CLASSES = [OIDCAuthentication]
+else:
+    AUTHENTICATION_CLASSES = [SessionAuthentication, BasicAuthentication]
+
 
 class AllocationViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Allocation.objects.filter(status__name='Active')
     serializer_class = serializers.AllocationSerializer
+    authentication_classes = AUTHENTICATION_CLASSES
+    permission_classes = [IsAdminUser]
 
 
 router = routers.SimpleRouter()
--- a/src/coldfront_plugin_cloud/urls.py
+++ b/src/coldfront_plugin_cloud/urls.py
@@ -1,0 +1,16 @@
+from rest_framework import routers, viewsets
+
+from coldfront.core.allocation.models import Allocation
+
+from coldfront_plugin_cloud.api import serializers
+
+
+class AllocationViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = Allocation.objects.filter(status__name='Active')
+    serializer_class = serializers.AllocationSerializer
+
+
+router = routers.SimpleRouter()
+router.register(r'allocations', AllocationViewSet)
+
+urlpatterns = router.urls
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,7 +20,8 @@ package_dir =
 packages = find:
 python_requires = >=3.8
 install_requires =
-    coldfront >= 1.0.4
+    coldfront @ git+https://github.com/knikolla/coldfront@coldfront_plugin_cloud_urls#egg=coldfront
+    djangorestframework
     python-cinderclient
     python-keystoneclient
     python-memcached == 1.59
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,6 +22,7 @@ python_requires = >=3.8
 install_requires =
     coldfront @ git+https://github.com/knikolla/coldfront@coldfront_plugin_cloud_urls#egg=coldfront
     djangorestframework
+    mozilla-django-oidc
     python-cinderclient
     python-keystoneclient
     python-memcached == 1.59
--- a/src/coldfront_plugin_cloud/api/serializers.py
+++ b/src/coldfront_plugin_cloud/api/serializers.py
@@ -1,0 +1,65 @@
+from rest_framework import serializers
+
+from coldfront.core.allocation.models import Allocation
+from coldfront.core.allocation.models import Project
+
+from coldfront_plugin_cloud import attributes
+
+
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = [
+            'id',
+            'title',
+            'pi',
+            'description',
+            'field_of_science',
+            'status'
+        ]
+
+    pi = serializers.SerializerMethodField()
+    field_of_science = serializers.SerializerMethodField()
+    status = serializers.SerializerMethodField()
+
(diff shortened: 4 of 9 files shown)


src/coldfront_plugin_cloud/urls.py
```diff
@@ -0,0 +1,53 @@
...
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = [
+            'title',
+            'pi',
+            'description',
+            'field_of_science',
+            'status'
+        ]
```
rob-baron, line 19: This seems to be a very specific structure that encodes our business logic as opposed to mapping our business logic to generic fields.

knikolla: Thank you for your feedback Rob,

I have found that wide formats are the defacto standard when defining schemas for REST APIs and that they are significantly more legible, easier to parse, and visualize.

As we're not trying to define a standard API interface that can accommodate other use cases (not related to ColdFront), in this instance I strongly prefer the representation encoding somewhat closely the ColdFront's database schema.

In this particular case for example, a resource allocation would be described as this.

```json
{
    "id": "123",
    "project": {
        "title": "This is the title of the project",
        "description": "this is the description of the project",
        "pi": "test@example.com",
        "field_of_science": "physical science",
        "status": "Active"
    },
    "description": "This is the desc of the resource allocation within the project",
    "resource": "NERC OpenShift",
    "status": "Active",
    "attributes": {
        "Allocated Project ID": "EA3B65EC-737C-4944-8B23-C4677F5EC4DA",
        "Allocated Project Name": "This is the project name in OpenStack"
    }
}
```

knikolla: Rob, nothing in this schema (outside of the `attributes` dictionary) is _our_ business logic and all of it is ColdFront's business logic. I am exactly structuring it like this for that purpose, to have something that is useful to ColdFront users and operators since it matches the representation of the model that ColdFront stores.

Operators that have nothing to do with us can consume a list of allocations, and view the ColdFront project information for that resource allocations. There is nothing specific about NERC, OpenStack, OpenShift, or anything of that matter in this part of the schema.

The only thing that is business logic, is all inside of the `attributes` field and this will be dependent on what the operator is attempting to do via the API and their specific setup.

In our case, we want to see the OpenStack/OpenShift project that is associated with a Resource Allocation in ColdFront. Per `coldfront-plugin-cloud` documentation, that field is stored in the `Allocated Project ID` and `Allocated Project Name` attributes. The Project sub structure is part of the upstream ColdFront schema, therefore it is generic and 100% not our business logic. 

ColdFront doesn't store any information about PIs and what their field of science is, it only stores that information about projects. The field `field_of_science` that you are referring to is inside of the `project` object, therefore it belongs to the project.

rob-baron: OK, I get what you are doing.  So I will have to report on the cold-front project name, converting either the openstack or openshift name/ID to the cold-front project when I am reporting on xdmod.  So the cold-front project name becomes an implicit additional layer to the hierarchy.

src/coldfront_plugin_cloud/urls.py
larsks, line 1: I'm not familiar with Django. Where/how is this file used? I don't see it imported anywhere in this project.

knikolla: This is the reason for the fork. This file is imported from `coldfront/config/urls.py`. https://github.com/knikolla/coldfront/commit/edf99cf3104ee96ad7d3d27dabb0b297921344ea

requirements.txt
```diff
@@ -1,4 +1,5 @@
-coldfront >= 1.0.4
+git+https://github.com/knikolla/coldfront@coldfront_plugin_cloud_urls#egg=coldfront
```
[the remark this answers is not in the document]
knikolla: I have commented about the need for a fork [here](https://github.com/ubccr/coldfront/pull/425#issuecomment-1503692098). Having a dynamic way to load `urls.py` seems like a sensible change to work with upstream to get included.

src/coldfront_plugin_cloud/urls.py
rob-baron: @knikolla Not to be condescending, but please think about what you stated and explain to me how a "Resource Allocation" is not a logical layer of the hierarchy.

knikolla: @rob-baron I'm not saying it is not a logical layer of the hierarchy. I'm saying that you are not required to introduce that layer for your reporting purposes.

It is your choice whether you want to report things as

```
OpenStack Project <--1 to 1--> PI, field_of_science
```

or 

```
OpenStack Project <--1 to 1--> Resource Allocation <--1 to many--> ColdFront Project <--1 to 1--> PI, field of science
```

All of that data is presented here and you are free to use it however you think best fulfills your use cases for reporting.

knikolla:
I have updated the PR with a new commit for authentication using OIDC. So if you have already reviewed the first commit, you can review just the second one in isolation.

src/coldfront_plugin_cloud/urls.py
larsks: @rob-baron I'm not sure I follow the concerns you've expressed here with the serialization format. I agree with @knikolla that representing the data on the wire in a format that closely matches the database representation is probably the least surprising decision, and at first glance I think it seems easier to consume. Can you present an example that demonstrates why this format is problematic?

rob-baron: @larsks - I am just trying to figure out the relationships as it is becoming lesss  and less clear to me.

I will just have to figure it out once the data comes in.

rob-baron (approved):

src/coldfront_plugin_cloud/urls.py
rob-baron: @knikolla 

Major conclusion first: let's get the data flowing!

Conclusions/concerns from our conversation:

It will be easier once I start processing the data as I strongly suspect it fits within an inheritance hierarchy and is not many-to-many.  This is the part that is becoming less clear, though I suspect we have been talking past each other. (for example people can be assigned many projects, and projects can have many people - this would be a many-to-many, where as 1 PI can run many projects and any project can have 1 PI  would be a 1-to-many).

I am expecting that the cold-front project name will need to be included with the openstack/openshift project name as that will make the specific of the project clearer to everyone.  I suspect that the PI will only know the projects by the cold-front project name, where as the technicians will know them by the openstack/openshift names.  This can be mapping is done on the xdmod side using the names.csv file, which enables any part of the hierarchy to be renamed.  Both of these will need to be reported as there is strong potential that everyone will become confused if only one is selected.

I am also concerned with inconsistencies between keycloak, cold-front, openshift and openstack.  It is why I am setting up a separate database just for tracking the changes in the heirarchy data so that I have a small chance of catching the inconsistencies as they come up (mostly manually).

knikolla: @rob-baron Thank you, this comment explained your concerns and confusion clearly.

joachimweyl:
Hello @knikolla,
Now that this has two approvals is it ready for merging?

jtriley (commented):
This feels out-of-scope for this plugin and would probably be better as a PR upstream given that this isn't really specific to provisioning OpenStack/OpenShift resources. Also the fork in the requirements is going to make things difficult for other folks that might be using the plugin without the need for this feature.

jtriley:
A better approach would be to integrate this into the fork (and also submit the feature upstream) and include that in coldfront-nerc for now until it gets merged upstream.

knikolla:
@jtriley @joachimweyl I have moved this to it's own repo since it was completely standalone code-wise and that sounds preferable to maintaining a fork.

joachimweyl:
@knikolla is [this](https://github.com/CCI-MOC/coldfront-plugin-api) the new repo? If so, does it close https://github.com/nerc-project/coldfront-plugin-cloud/issues/89

PR #95, opened 2023-04-14
fix permission issue setting object-store quotas with radosgw swift

jtriley:


src/coldfront_plugin_cloud/openstack.py
```diff
@@ -157,21 +176,63 @@ def set_quota(self, project_id):
...
+                f'OPENSTACK_{var_name}_APPLICATION_CREDENTIAL_SECRET'
+            ).encode('utf-8')
+        )
+
+        password = phash.hexdigest()[0:int(phash.block_size/2)]
+
+        try:
+            user = self.identity.users.create(
+                name=COLDFRONT_RGW_SWIFT_INIT_USER,
+                project=project_id,
```
knikolla, line 214: This doesn't grant any actual authorization on that specific project and only specifies the *default* project of the user. You have to make an additional call to assign the user a role on the project.

From the [Keystone API documentation](https://docs.openstack.org/api-ref/identity/v3/?expanded=create-user-detail#create-user)

```The ID of the default project for the user. A user’s default project must not be a domain. Setting this attribute does not grant any actual authorization on the project, and is merely provided for convenience. Therefore, the referenced project does not need to exist within the user domain. (Since v3.1) If the user does not have authorization to their default project, the default project is ignored at token creation. (Since v3.1) Additionally, if your default project is not valid, a token is issued without an explicit scope of authorization.```

knikolla (changes requested):

src/coldfront_plugin_cloud/openstack.py
jtriley: Thanks for the pointer! I've updated the code to remove the `project_id` kwarg from the user create call. I've also added and`self.assign_role_on_user()` and then `self.remove_role_from_user()` after the `head_account()` returns successfully in `self._init_rgw_for_project()`.

I've also removed the silent logging in case of swift errors. I'm debating whether to also cleanup the `coldfront-swift-init` user or just leave it around given that each new project will need it to setup quotas.

jtriley:
> Just noting this failure which is what this PR is attempting to fix: https://github.com/nerc-project/coldfront-plugin-cloud/actions/runs/4702343428/jobs/8339479639?pr=95
> 
> ```
> ======================================================================
> ERROR: test_new_allocation (coldfront_plugin_cloud.tests.functional.openstack.test_allocation.TestAllocation)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/runner/work/coldfront-plugin-cloud/coldfront-plugin-cloud/src/coldfront_plugin_cloud/tests/functional/openstack/test_allocation.py", line 104, in test_new_allocation
>     resulting_quota = allocator.get_quota(openstack_project.id)
>   File "/home/runner/work/coldfront-plugin-cloud/coldfront-plugin-cloud/src/coldfront_plugin_cloud/openstack.py", line 249, in get_quota
>     swift = self.object(project_id).head_account()
>   File "/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/swiftclient/client.py", line 1850, in head_account
>     return self._retry(None, head_account, headers=headers)
>   File "/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/swiftclient/client.py", line 1797, in _retry
>     rv = func(self.url, self.token, *args,
>   File "/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/swiftclient/client.py", line 870, in head_account
>     raise ClientException.from_response(resp, 'Account HEAD failed', body)
> swiftclient.exceptions.ClientException: Account HEAD failed: http://10.1.0.94/swift/v1/AUTH_4434b4126e7642e9b6310357790e9b58 401 Unauthorized (txn: tx00000f3141a1e8d30f3c9-006439900e-3796-default)
> ```

This is actually a different, but similar, exception than what this PR is attempting to fix. This PR attempts to fix the 403 Forbidden case. I believe the above 401 error was due to the keystone auth from the radosgw failing due to SSL verification.

knikolla (approved):
Seems good to me!

PR #98, opened 2023-05-05
add management command to list cloud allocations

jtriley:
This adds a new django admin management command: `list_cloud_allocations`. This prints a list of all cloud allocations (ie type openstack or openshift) and their attributes in either json (default) or csv format.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/list_cloud_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/list_cloud_allocations.py
@@ -1,0 +1,124 @@
+import sys
+import csv
+import json
+import logging
+
+from django.core.management.base import BaseCommand
+from django.core.exceptions import ObjectDoesNotExist
+
+from coldfront_plugin_cloud import attributes
+from coldfront_plugin_cloud import openstack
+from coldfront.core.resource.models import (Resource, ResourceType)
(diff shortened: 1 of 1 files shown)


knikolla (approved):

src/coldfront_plugin_cloud/management/commands/list_cloud_allocations.py
```diff
@@ -0,0 +1,124 @@
...
+        parser.add_argument(
+            '--format',
+            choices=['json', 'csv'],
+            default='json',
+            help='output format'
+        )
+
+    def get_cloud_attrs(self, cloud_type):
+        attrs = [
+            i for i in attributes.ALLOCATION_QUOTA_ATTRIBUTES if cloud_type in i
```
knikolla, line 41: This will only get you the allocation attributes, if you want to get all attributes I'm using this https://github.com/CCI-MOC/coldfront-plugin-api/blob/2ccc5679b9d9cd407ef939425e19929cbc0d3fd0/src/coldfront_plugin_api/serializers.py#L57-L61

joachimweyl: @jtriley what are the next steps? What is the priority for this PR?

PR #117, opened 2023-10-02
Change headers and add cost calculation

knikolla:


src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -24,28 +25,40 @@ class InvoiceRow:
...
+    Cost: Decimal = 0
+
 
     @classmethod
     def get_headers(cls):
-        """Returns all fields for display, with space instead of underscore."""
+        """Returns all headers for display."""
         return [
-            field.name.replace("_", " ") for field in dataclasses.fields(cls)
+            "Interval",
```
joachimweyl, line 38: Invoice Month

knikolla: The tool accepts arbitrary start and end dates that don't have to conform to an exact month, therefore with the current logic of the code the header Invoice Month would be incorrect.

joachimweyl: That is the name of the Header as it houses only a specific month (eg. 2023-09). The value of the field in the code might hold something else during the processing of that output but the header needs to match the header of the CSV we are sending to MGHPCC.

knikolla: Yes, right now I am changing the header and entire column manually before sending it to you. I will change the tool to output the header that you expect (Invoice Month) in a follow up patch as that requires adding new parameters to the execution of the tool.

joachimweyl (approved):

knikolla:
Job `test-py39-functional-microshift` was slightly flaky. Re-running it fixed the issue.

PR #119, opened 2023-10-31
updated the new default value for 1 Unit of resource allocation

Milstein:


joachimweyl (approved):

knikolla:
@Milstein Looks like some of the tests held some assumptions with regards to the default quotas. Can you please update the tests?

joachimweyl:
@knikolla any thoughts on why it failed the py39-functional test?

knikolla:
@joachimweyl `py39-functional` installs Ceph and RGW to test setting swift quotas. In this case the Ceph installation is failing.

knikolla (approved):

larsks:
@knikolla do you agree that this failure seems to be outside of our control, so we should just ignore it (and possibly bug report it upstream)?

knikolla:
> @knikolla do you agree that this failure seems to be outside of our control, so we should just ignore it (and possibly bug report it upstream)?

Yes, I'm okay with merging this despite CI failures for now while we work on a fix for the CI. (#122)

PR #122, opened 2023-11-06
fix recent radosgw CI/CD pipeline failures

jtriley:
This fixes an issue with the radosgw portion of the CI/CD pipeline due to an upstream issue/change with ceph-exporter related to the metrics setup on the cluster. This patch simply disables setting up the monitoring stack altogether given that the metrics setup is not used by this pipeline.

This patch also doubles the size of the OSDs given that the most recent version of ceph deployed by cephadm 17.2.6-0ubuntu0.22.04.1 requires more space. This was determined by the appearance of out-of-space errors
in OSD logs on an external test VM.

In addition to these changes, this patch also disables configuring firewalld given that it's not installed and not necessary for the CI/CD pipeline.

### DIFF
--- a/ci/radosgw.sh
+++ b/ci/radosgw.sh
@@ -42,18 +42,19 @@ EOF
     --config /tmp/ceph.conf \
     --cluster-network $NETWORK \
     --mon-ip $IP \
-    --dashboard-password-noupdate \
-    --initial-dashboard-user admin \
-    --initial-dashboard-password ceph \
     --allow-fqdn-hostname \
-    --single-host-defaults
+    --single-host-defaults \
+    --log-to-file \
+    --skip-firewalld \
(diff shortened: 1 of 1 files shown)


jtriley:
Testing this locally in a ubuntu 22.04 VM is showing the rgw container failing with:

```
debug 2023-11-07T15:57:21.775+0000 7fb059c64700 -1 Initialization timeout, failed to initialize
```

Still looking into it.

jtriley:
Tracked this down to an issue with OSDs not successfully joining the cluster due to out of space issue during initialization. Trying to double the size of the OSDs (from 500M to 1000M) to see if that helps.

aabaris (approved):

knikolla (approved):

PR #134, opened 2024-01-17
Merging the account manager into the plugin

### LINKED ISSUE #137: Copy over the code from `openshift-acct-mgt` with minimum changes
The initial version should just copy the entire package and do minimum changes to get the function calls and unit/functional testing to work.

Motivation #114

QuanMPhm:
Closes #137, I am merging the account manager into the plugin. Most changes are in `openshift.py`


src/coldfront_plugin_cloud/openshift.py
```diff
@@ -33,75 +111,76 @@ class Conflict(ApiException):
...
+            logger = logging.getLogger("django")
+
+        config = env_config()
+
+        self.client = moc_openshift.MocOpenShift4x(
+            DynamicClient(k8s_client), logger, config, QUOTA_OPENSHIFT, LIMITS_OPENSHIFT
+        )
+
     @functools.cached_property
     def session(self):
```
knikolla, line 140: Session is not required anymore since you're not making API calls to the `acct-mgt` service through HTTPS.

src/coldfront_plugin_cloud/acct_mgt/defaults.py
```diff
@@ -0,0 +1,5 @@
+"""Default values for Flask app configuration"""
```
knikolla, line 1: This file is not required anymore.

src/coldfront_plugin_cloud/acct_mgt/moc_openshift.py
```diff
@@ -0,0 +1,462 @@
...
+            rolebinding = self.get_rolebindings(project_name, role)
+
+            if not self.user_in_rolebinding(user_name, rolebinding):
+                rolebinding["subjects"].append({"kind": "User", "name": user_name})
+                self.update_rolebindings(project_name, rolebinding)
+        except kexc.NotFoundError:
+            rolebinding = self.create_rolebindings(project_name, user_name, role)
+
+        return {
+            "msg": f"added user {user_name} to role {role} in {project_name}",
```
knikolla, line 116: This comment applies to the entire file. Log these messages instead of returning them.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -33,75 +111,76 @@ class Conflict(ApiException):
...
 
     project_name_max_length = 63
 
+    def __init__(
+        self,
+        resource: resource_models.Resource,
+        allocation: allocation_models.Allocation,
+    ):
+        self.resource = resource
+        self.allocation = allocation
```
knikolla, line 124: Since these are operations performed by the base class, I suggest letting the base class handle them and calling the base class constructor. See https://realpython.com/python-super/

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -113,70 +192,116 @@ def reactivate_project(self, project_id):
...
+                self.client.create_identity(id_user)
+
+            if not self.client.useridentitymapping_exists(unique_id, id_user):
+                created = True
+                self.client.create_useridentitymapping(unique_id, id_user)
+
+                return "TODO create user"
+
+            if created:
+                return {"msg": f"user created ({unique_id})"}
```
knikolla, line 225: These should be logged rather than returned.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -113,70 +192,116 @@ def reactivate_project(self, project_id):
             pass
 
     def get_federated_user(self, username):
-        url = f"{self.auth_url}/users/{username}"
         try:
-            r = self.session.get(url)
-            self.check_response(r)
-            return {'username': username}
+            if self.client.user_exists(username):
```
knikolla, line 196: A user existing should mean all components of the user. The `user` the `identity` and the `useridentitymapping`.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -7,8 +7,71 @@
...
 
+import kubernetes.config
+import kubernetes.client
+import kubernetes.dynamic.exceptions as kexc
+from openshift.dynamic import DynamicClient
+
+from coldfront.core.allocation import models as allocation_models
+from coldfront.core.resource import models as resource_models
+
+from .acct_mgt import moc_openshift
```
knikolla, line 18: Please use the same syntax for importing the module as the rest of the imports (absolute rather than relative.)

In addition, generally the ordering of imports that I have followed in this repository is
```
python standard library imports
empty line
python imports from library not in standard library
empty line
python imports from the same package as this one (so everything from `coldfront_plugin_cloud`)
```

Please try to follow the same conventions when possible.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -18,6 +81,21 @@
     attributes.QUOTA_PVC: lambda x: {":persistentvolumeclaims": f"{x}"},
 }
 
+ENVPREFIX = "ACCT_MGT_"
```
knikolla, line 84: These shouldn't be "ACCT_MGT_" anymore.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -18,6 +81,21 @@
...
+
+
+def env_config():
```
knikolla, line 87: It should be possible to load different configurations in a scenario where there are multiple OpenShift Resources.(Ex. see how you can load different credentials based on the Resource name as we did earlier for both OpenStack and OpenShift)

knikolla (changes requested):

larsks (changes requested):
I would like to see this series of commits cleaned up:

- Each commit should have a clear purpose and should be clearly described. There should be a short descriptive subject followed by information about why the changes are necessary.
- There should not be multiple commits with the same subject.
- Please check messages for spelling and grammar.
- Typo fixes and other cleanups should be folded into other commits.

QuanMPhm:
I will likely need to close this PR and open a new one, hopefully with a lot more cleanup and up to standard

larsks:
> I will likely need to close this PR and open a new one

Please don't do that. Just update this PR.

knikolla:
Great work on cleaning up the commits. One final advice. You can see that when you merged the main branch back into this working branch, it introduced a series of extraneous commits into your history. It's generally better to rebase your working branch into the main branch. Please look into and remove the merge commits from your history. I suggest you backup your work too, just in case something goes wrong modifying the history.

See this Stackoverflow answer quickly describing the differences between the two workflows https://stackoverflow.com/questions/804115/when-do-you-use-git-rebase-instead-of-git-merge

QuanMPhm:
@knikolla I've resolved all comments so far. The things I see that are left to do is to rename the OS_AUTH_URL attribute, and update the test files as you see fit. I will wait for further instructions.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -33,75 +83,74 @@ class Conflict(ApiException):
...
-            except JSONDecodeError:
-                # https://github.com/CCI-MOC/openshift-acct-mgt/issues/54
-                return response.text
-        if response.status_code == 404:
-            raise NotFound(f"{response.status_code}: {response.text}")
-        elif 'does not exist' in response.text or 'not found' in response.text:
-            raise NotFound(f"{response.status_code}: {response.text}")
-        elif 'already exists' in response.text:
-            raise Conflict(f"{response.status_code}: {response.text}")
+        openshift_token = os.getenv(f"OPENSHIFT_{var_name}_TOKEN")
```
larsks, line 99: Instead of passing in a token, can we just read the configuration from a `kubeconfig` file? Using `config.load_kubeconfig()` will work with both the in-cluster configuration for a serviceaccount and with a local `kubeconfig` file (and respects the `KUBECONFIG` environment variable).

If you don't want to change this now, it may be worth turning into an issue.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -113,70 +162,120 @@ def reactivate_project(self, project_id):
...
+
+            if not self.client.identity_exists(id_user):
+                created = True
+                self.client.create_identity(id_user)
+
+            if not self.client.useridentitymapping_exists(unique_id, id_user):
+                created = True
+                self.client.create_useridentitymapping(unique_id, id_user)
+
+                return "TODO create user"
```
larsks, line 193: Why are we returning a string here (and nowhere else in this function)?

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -113,70 +162,120 @@ def reactivate_project(self, project_id):
...
+
+            if created:
+                self.logger.info(f"msg: user created ({unique_id})")
+                return
+
+            raise Conflict("400: " + f"user already exists ({unique_id})")
```
larsks, line 199: Avoid including http response codes in our text messages. We're already providing the response code to the client as part of the HTTP response. Also, we should be returning a 409 status code on conflicts.

QuanMPhm: @larsks @knikolla  Currently I am handling the exceptions the same way that the [current code](https://github.com/nerc-project/coldfront-plugin-cloud/blob/1fff22e096c3440bb783f1152a985f14eeef47b9/src/coldfront_plugin_cloud/openshift.py#L67) is handling it, by passing the response and the message. Should I still remove all error codes?

As for whether the `Conflict` exception returns the appropriate code, that doesn't seem to be implemented yet. `Conflict` is simply subclassed from the built-in `Exception` class. Let me know if I should implement it.

QuanMPhm: From what I've read about Django exception handling, there is no way to raise custom exceptions in Django which returns custom status codes in the http response, especially not in a signal handler, as is the case with `openshift.py`. 
When an error is raised (as in the example you highlighted), Django by default returns `500` as the status code.

The only way I can see to return appropriate errors is by including it in the error message. Please tell me if I missed something and that there is a clean way of returning custom errors.

src/coldfront_plugin_cloud/openshift.py
knikolla: For now we can keep passing the token as an environment variable, as that preserves the original behavior. We can work out a better way in the future.

knikolla (changes requested):
I'll do another pass later today as I ran out of time now.

src/coldfront_plugin_cloud/openshift.py
knikolla: We don't need to return custom error codes in Django. These exceptions are purely for the code itself to handle, not any external user.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -113,70 +161,118 @@ def reactivate_project(self, project_id):
...
+            + f"user role does not exist ({project_id},{username},{self.member_role_name})"
+        )
 
     def _get_project(self, project_id):
-        url = f"{self.auth_url}/projects/{project_id}"
-        r = self.session.get(url)
-        return self.check_response(r)
+        if self.client.project_exists(project_id):
+            self.logger.info(f"msg: project exists ({project_id})")
+            return
```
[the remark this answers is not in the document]
QuanMPhm: It seems the purpose of `_get_role` and `_get_project` is [purely to check i](https://github.com/nerc-project/coldfront-plugin-cloud/blob/6228541b18ca776124d5096ff8ecca2109df1634/src/coldfront_plugin_cloud/tests/functional/openshift/test_allocation.py#L53)f the role or project has been created on the Openshift cluster. I've actually broke the test cases here by converting the exception raise to a log message.

joachimweyl:
Blocked on Functional test.

QuanMPhm:
@knikolla I have brought over the unit tests from the account manager and hopefully resolved all problems with the functional tests

src/coldfront_plugin_cloud/openshift.py
joachimweyl: @QuanMPhm if you have made this change please resolve this conversation or ask one of those in the conversation to resolve it if you are not sure.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -113,70 +161,118 @@ def reactivate_project(self, project_id):
...
+
+            if not self.client.useridentitymapping_exists(unique_id, id_user):
+                created = True
+                self.client.create_useridentitymapping(unique_id, id_user)
+
+            if created:
+                self.logger.info(f"msg: user created ({unique_id})")
+                return
+
+            raise Conflict("409: " + f"user already exists ({unique_id})")
```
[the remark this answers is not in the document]
joachimweyl: @QuanMPhm if you removed the line please resolve this conversation.

src/coldfront_plugin_cloud/openshift.py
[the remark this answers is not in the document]
joachimweyl: @knikolla what are your thoughts on this approach?

joachimweyl:
@QuanMPhm what are the next steps?

joachimweyl:
@QuanMPhm what are the next steps?

QuanMPhm:
Now that this PR has been refactored into smaller PRs (#175 #187 #186 #185 #184), I will be closing this now.

PR #146, opened 2024-03-11
Integrate ESI into Coldfront

### LINKED ISSUE #135: Support for ESI
https://github.com/CCI-MOC/esi/issues/364

> Most things should work out of the box as ESI is OpenStack and there's support for OpenStack via the [coldfront cloud plugin](https://github.com/nerc-project/coldfront-plugin-cloud)
>
> The thing we will need to figure out is quotas for things that are specific to ESI e.g. VLAN IDs and anything else that ESI does outside of OpenStack (if any).

This is the ColdFront side of ensuring that Coldfront can support ESI. The ESI side of ensuring this communiation works can be found [here](https://github.com/CCI-MOC/esi/issues/364).

QuanMPhm:
Closes #135 To integrate ESI and make use of the existing code, I have made a few small restructuring choices. There’s too many to list, so I will be ready for everyone’s questions on my code changes. 

The ESI allocator object, implemented in `esi.py`, subclasses from the Openstack allocator, with a few functions overridden. The ESI allocator only makes quotas for floating IPs and networks. The appropriate Allocation Attributes have been added to `attributes.py`

The functional test for the ESI integration uses the same Openstack cluster put up by `test-py38-functional-devstack.yaml`. `ci/run_functional_tests_openstack.sh` has been edited to also create an ESI coldfront resource and to run the ESI functional tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,11 +78,14 @@ dashboard or through the helper command:
 
 ```bash
 $ coldfront add_openstack_resource
-usage: coldfront add_openstack_resource [-h] --name NAME --auth-url AUTH_URL [--users-domain USERS_DOMAIN] [--projects-domain PROJECTS_DOMAIN] --idp IDP
-                                        [--protocol PROTOCOL] [--role ROLE] [--version] [-v {0,1,2,3}] [--settings SETTINGS] [--pythonpath PYTHONPATH] [--traceback]
-                                        [--no-color] [--force-color]
+usage: coldfront add_openstack_resource [-h] --name NAME --auth-url AUTH_URL [--users-domain USERS_DOMAIN] [--projects-domain PROJECTS_DOMAIN] --idp IDP [--protocol PROTOCOL] [--role ROLE]
+                                        [--public-network PUBLIC_NETWORK] [--network-cidr NETWORK_CIDR] [--esi] [--version] [-v {0,1,2,3}] [--settings SETTINGS] [--pythonpath PYTHONPATH] [--traceback]
+                                        [--no-color] [--force-color] [--skip-checks]
 coldfront add_openstack_resource: error: the following arguments are required: --name, --auth-url, --idp
 ```
+
+An Openstack resource can be specified as an ESI resource by setting the `--esi` command flag.
+ESI resource allocations will only have quotas for network resources by default.
 
 ### Configuring for OpenShift
 
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -75,6 +75,7 @@ QUOTA_VOLUMES = 'OpenStack Number of Volumes Quota'
 QUOTA_VOLUMES_GB = 'OpenStack Volume Quota (GiB)'
 
 QUOTA_FLOATING_IPS = 'OpenStack Floating IP Quota'
+QUOTA_NETWORKS = 'Openstack Network Quota'
 
 QUOTA_OBJECT_GB = 'OpenStack Swift Quota (GiB)'
 
@@ -96,6 +97,7 @@ ALLOCATION_QUOTA_ATTRIBUTES = [
     CloudAllocationAttribute(name=QUOTA_VCPU),
     CloudAllocationAttribute(name=QUOTA_VOLUMES),
     CloudAllocationAttribute(name=QUOTA_VOLUMES_GB),
+    CloudAllocationAttribute(name=QUOTA_NETWORKS),
     CloudAllocationAttribute(name=QUOTA_FLOATING_IPS),
     CloudAllocationAttribute(name=QUOTA_OBJECT_GB),
     CloudAllocationAttribute(name=QUOTA_GPU),
--- a/src/coldfront_plugin_cloud/esi.py
+++ b/src/coldfront_plugin_cloud/esi.py
@@ -1,0 +1,29 @@
+import logging
+import functools
+import os
+
+from keystoneauth1.identity import v3
+from keystoneauth1 import session
+
+from coldfront_plugin_cloud import attributes, utils
+from coldfront_plugin_cloud.openstack import OpenStackResourceAllocator
+
+class ESIResourceAllocator(OpenStackResourceAllocator):
+
+    QUOTA_KEY_MAPPING = {
+        'network': {
+            'keys': {
+                attributes.QUOTA_FLOATING_IPS: 'floatingip',
+                attributes.QUOTA_NETWORKS: 'network'
+            }
+        }
+    }
+
+    QUOTA_KEY_MAPPING_ALL_KEYS = {quota_key: quota_name for k in QUOTA_KEY_MAPPING.values() for quota_key, quota_name in k['keys'].items()}
+
+    resource_type = 'esi'
+
+    def get_quota(self, project_id):
+        quotas = dict()
+        quotas = self._get_network_quota(quotas, project_id)
+        return quotas
--- a/src/coldfront_plugin_cloud/management/commands/add_openstack_resource.py
+++ b/src/coldfront_plugin_cloud/management/commands/add_openstack_resource.py
@@ -34,13 +34,23 @@ class Command(BaseCommand):
         parser.add_argument('--network-cidr', type=str, default='192.168.0.0/24',
                             help='CIDR for default networks. '
                                  'Ignored if no --public-network.')
+        parser.add_argument('--esi', action='store_true',
+                            help='Indicates this is an ESI resource (default: False)')
 
     def handle(self, *args, **options):
+
+        if options['esi']:
+            resource_description = 'ESI Bare Metal environment'
+            resource_type = 'ESI'
+        else:
+            resource_description = 'OpenStack cloud environment'
+            resource_type = 'OpenStack'
+
         openstack, _ = Resource.objects.get_or_create(
-            resource_type=ResourceType.objects.get(name='OpenStack'),
+            resource_type=ResourceType.objects.get(name=resource_type),
             parent_resource=None,
             name=options['name'],
-            description='OpenStack cloud environment',
+            description=resource_description,
             is_available=True,
             is_public=True,
             is_allocatable=True
@@ -82,18 +92,21 @@ class Command(BaseCommand):
             resource=openstack,
             value=options['role']
         )
-        ResourceAttribute.objects.get_or_create(
-            resource_attribute_type=ResourceAttributeType.objects.get(
-                name='quantity_label'),
-            resource=openstack,
-            value='Units of computing to allocate to the project. 1 Unit = 1 Instance, 2 vCPU, 0 GPU, 4G RAM, 2 Volumes, 100 GB Volume Storage, and 1 GB Object Storage'
-        )
-        ResourceAttribute.objects.get_or_create(
-            resource_attribute_type=ResourceAttributeType.objects.get(
-                name='quantity_default_value'),
-            resource=openstack,
-            value=1
-        )
+
+        # Quantity values do not make sense for an ESI allocation
+        if not options['esi']:
+            ResourceAttribute.objects.get_or_create(
+                resource_attribute_type=ResourceAttributeType.objects.get(
+                    name='quantity_label'),
+                resource=openstack,
+                value='Units of computing to allocate to the project. 1 Unit = 1 Instance, 2 vCPU, 0 GPU, 4G RAM, 2 Volumes, 100 GB Volume Storage, and 1 GB Object Storage'
+            )
+            ResourceAttribute.objects.get_or_create(
+                resource_attribute_type=ResourceAttributeType.objects.get(
+                    name='quantity_default_value'),
+                resource=openstack,
+                value=1
+            )
 
         if options['public_network']:
             ResourceAttribute.objects.get_or_create(
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -4,6 +4,7 @@ import re
 from coldfront_plugin_cloud import attributes
 from coldfront_plugin_cloud import openstack
 from coldfront_plugin_cloud import openshift
+from coldfront_plugin_cloud import esi
 from coldfront_plugin_cloud import utils
 from coldfront_plugin_cloud import tasks
 
@@ -66,11 +67,9 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
 
-        # Openstack Resources first
+        # Deal with Openstack and ESI resources
         openstack_resources = Resource.objects.filter(
-            resource_type=ResourceType.objects.get(
-                name='OpenStack'
-            )
+            resource_type__name__in=['OpenStack', 'ESI']
         )
         openstack_allocations = Allocation.objects.filter(
             resources__in=openstack_resources,
@@ -84,10 +83,7 @@ class Command(BaseCommand):
 
             failed_validation = False
 
-            allocator = openstack.OpenStackResourceAllocator(
-                allocation.resources.first(),
-                allocation
-            )
+            allocator = tasks.find_allocator(allocation)
 
             project_id = allocation.get_attribute(attributes.ALLOCATION_PROJECT_ID)
             if not project_id:
@@ -105,11 +101,11 @@ class Command(BaseCommand):
 
             failed_validation = Command.sync_users(project_id, allocation, allocator, options["apply"])
 
-            obj_key = openstack.QUOTA_KEY_MAPPING['object']['keys'][attributes.QUOTA_OBJECT_GB]
+            obj_key = openstack.OpenStackResourceAllocator.QUOTA_KEY_MAPPING['object']['keys'][attributes.QUOTA_OBJECT_GB]
 
             for attr in attributes.ALLOCATION_QUOTA_ATTRIBUTES:
                 if 'OpenStack' in attr.name:
-                    key = openstack.QUOTA_KEY_MAPPING_ALL_KEYS.get(attr.name, None)
+                    key = allocator.QUOTA_KEY_MAPPING_ALL_KEYS.get(attr.name, None)
                     if not key:
                         # Note(knikolla): Some attributes are only maintained
                         # for bookkeeping purposes and do not have a
@@ -147,7 +143,7 @@ class Command(BaseCommand):
                         allocation.get_attribute(attributes.ALLOCATION_PROJECT_ID)
                     )
                 except Exception as e:
-                    logger.error(f'setting openstack quota failed: {e}')
+                    logger.error(f'setting {allocation.resources.first()} quota failed: {e}')
                     continue
                 logger.warning(f'Quota for allocation {allocation_str} was out of date. Reapplied!')
 
--- a/src/coldfront_plugin_cloud/openstack.py
+++ b/src/coldfront_plugin_cloud/openstack.py
@@ -21,39 +21,7 @@ logger = logging.getLogger(__name__)
 # 1 GB = 1 000 000 000 B = 10^9 B
 GB_IN_BYTES = 2 ** 30
 
-# Map the attribute name in ColdFront, to the client of the respective
-# service, the version of the API, and the key in the payload.
-QUOTA_KEY_MAPPING = {
-    'compute': {
-        'keys': {
-            attributes.QUOTA_INSTANCES: 'instances',
-            attributes.QUOTA_VCPU: 'cores',
-            attributes.QUOTA_RAM: 'ram',
-        },
-    },
-    'network': {
-        'keys': {
-            attributes.QUOTA_FLOATING_IPS: 'floatingip',
-        }
-    },
-    'object': {
-        'keys': {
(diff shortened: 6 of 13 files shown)


tzumainn (commented):
Hi! Just a general comment: ESI doesn't require nova/compute, so that's something you'll want to take out of the ESI quotas (and probably the "small openstack" tests).

src/coldfront_plugin_cloud/tests/functional/openstack_mini/test_allocation.py
```diff
@@ -0,0 +1,55 @@
...
+
+from django.core.management import call_command
+from keystoneclient.v3 import client
+from cinderclient import client as cinderclient
+from neutronclient.v2_0 import client as neutronclient
+from novaclient import client as novaclient
+
+
+@unittest.skipUnless(os.getenv('FUNCTIONAL_TESTS'), 'Functional tests not enabled.')
+class TestSmallOpenstack(base.TestBase):
```
tzumainn, line 16: Maybe "BareMetalOpenStack"? And include identity/networking/volume?

QuanMPhm: That's not actually the unit test to test the ESI integration. That's just the test to make sure that Coldfront doesn't crashes if we attempt to make **Openstack** allocations on Openstack clusters that might be missing some services, like Cinder and Swift.

This is why in `ci/devstack_mini.sh`, I disabled cinder and swift.

QuanMPhm:
@tzumainn Do you have an answer for the other questions I mentioned in my PR comments? I feel like there is a lot more information I need to know before being able to remotely continue solving this issue.

tzumainn:
> @tzumainn Do you have an answer for the other questions I mentioned in my PR comments? I feel like there is a lot more information I need to know before being able to remotely continue solving this issue.

Whoops, sorry, I missed that there were question in the files themselves; in the future it may be easier to add questions as git comments on the PR and tag the relevant people.

Anyway, floating IPs and networks are the only items I can think of that might require allocations, since I think we explicitly do not want ColdFront to limit capacity for leasing nodes.

README.md
```diff
@@ -111,6 +111,9 @@ usage: coldfront add_openshift_resource [-h] --name NAME --auth-url AUTH_URL [--
 coldfront add_openshift_resource: error: the following arguments are required: --name, --auth-url
 ```
 
+### Configuring for ESI
+(Quan Pham) TODO Add instructions for configuring ESI
```
QuanMPhm, line 115: Are there any specific instructions for me as I update the README to include information about configuring ESI?

ci/devstack_esi.sh
```diff
@@ -0,0 +1,2 @@
+# (Quan Pham) TODO Know we want to install in a test Devstack cluster
+# to test ESI
```
QuanMPhm, line 2: What openstack services do we need for to deploy a Devstack cluster for the purpose of testing the ESI service? Do we want to test the ESI integration against more than one Devstack cluster?

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -89,6 +89,12 @@ class CloudAllocationAttribute:
 QUOTA_REQUESTS_GPU = 'OpenShift Request on GPU Quota'
 QUOTA_PVC = 'OpenShift Persistent Volume Claims Quota'
 
+###########################################################
+# ESI Quota Attributes 
+# (Quan Pham) TODO Know what ESI specific attributes would be, if any
+# From Mainn's suggesting, I've added networks and floating ips
```
QuanMPhm, line 95: What ESI specific allocation attribute should be included?

README.md
tzumainn: Ah, this I wouldn't know; I guess it would be `add_esi_resource`, since this would be distinct from OpenStack? Kristi may know for sure, though.

src/coldfront_plugin_cloud/esi.py
```diff
@@ -0,0 +1,66 @@
...
+from keystoneclient.v3 import client as ks_client
+from cinderclient import client as cinderclient
+from neutronclient.v2_0 import client as neutronclient
+from novaclient import client as novaclient
+
+from coldfront_plugin_cloud import attributes, base, utils
+
+from coldfront_plugin_cloud.openstack import OpenStackResourceAllocator
+
+# (Quan Pham) TODO Know which resources ESI needs, and their multipliers
```
QuanMPhm, line 21: What are the multipliers and static additions for the different ESI resources? And what exactly will these resources be?

src/coldfront_plugin_cloud/esi.py
```diff
@@ -0,0 +1,66 @@
...
+        auth,
+        verify=os.environ.get('FUNCTIONAL_TESTS', '') != 'True'
+    )
+
+class ESIResourceAllocator(OpenStackResourceAllocator):
+
+    resource_type = 'esi'
+
+    # (Quan Pham) TODO implement setting quotas for ESI resource
+    # And know which functions needs to be implement specially for ESI
```
QuanMPhm, line 58: This question can be addressed after we know exactly how Coldfront should interact with an ESI cluster when an ESI allocation is approved/denied: What are all the functions that should be implement for the ESI allocator?

src/coldfront_plugin_cloud/management/commands/add_esi_resource.py
```diff
@@ -0,0 +1,112 @@
...
+            resource_attribute_type=ResourceAttributeType.objects.get(
+                name=attributes.RESOURCE_ROLE),
+            resource=openstack,
+            value=options['role']
+        )
+        ResourceAttribute.objects.get_or_create(
+            resource_attribute_type=ResourceAttributeType.objects.get(
+                name='quantity_label'),
+            resource=openstack,
+            value='Unit of ???' # (Quan Pham) TODO Know what each allocation unit corresponds to
```
QuanMPhm, line 91: What should be the description text for an ESI allocation unit?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -253,3 +253,16 @@ def handle(self, *args, **options):
                             logger.warning(
                                 f"Quota for allocation {project_id} was out of date. Reapplied!"
                             )
+
+        # (Quan Pham) TODO Implement validating ESI resource
+        # Should these validation code be put in a sepearate file or 
+        # in tasks.py or utils.py as functions for each resource type?
```
QuanMPhm, line 259: How exactly should we validate ESI allocations? Since the test Devstack clusters we're running to test ESI integration might not require Ironic or ESI to run on them, how should the allocations be validated without directly testing ESI functionality?

src/coldfront_plugin_cloud/openstack.py
```diff
@@ -23,6 +23,13 @@
 
 # Map the attribute name in ColdFront, to the client of the respective
 # service, the version of the API, and the key in the payload.
+#
+# (Quan Pham) TODO: The name of the services should be the same as the
+# names or aliases listed in the Openstack Service Type Authority list 
+# (https://opendev.org/openstack/service-types-authority/)
+#
+# This is because we query the Openstack cluster to check what services 
+# are available. It is probably a bad idea to hardcode the service names here
```
QuanMPhm, line 32: This is less a question, and more of a proposal for a future issue. The names of the services listed in the `QUOTA_KEY_MAPPING` dict should match what is in the Openstack Service Type Authority to allow easier tracking of what Openstack objects/services our Coldfront plugin is directly interacting with.

src/coldfront_plugin_cloud/tests/functional/esi/test_allocations.py
```diff
@@ -0,0 +1,32 @@
...
+from keystoneclient.v3 import client
+from cinderclient import client as cinderclient
+from neutronclient.v2_0 import client as neutronclient
+from novaclient import client as novaclient
+
+@unittest.skipUnless(os.getenv('FUNCTIONAL_TESTS'), 'Functional tests not enabled.')
+class TestAllocation(base.TestBase):
+
+    # (Quan Pham) TODO What are the test cases for an ESI resource?
+    # What do we need to test?
```
QuanMPhm, line 18: How should the test cases for the ESI integration be designed?

ci/devstack_esi.sh
tzumainn: It depends what you're testing; ideally Keystone/Neutron/Ironic/Glance. However, if you're only testing quotas, you really only need Keystone/Neutron.

src/coldfront_plugin_cloud/attributes.py
tzumainn: Networks and floating IPs are all I can think of.

src/coldfront_plugin_cloud/esi.py
tzumainn: I'm not exactly sure what multipliers and static additions mean, so this may be a question for Kristi.

tzumainn:
I've answered the ESI specific questions; however, I'm not all that familiar with ColdFront itself, so this questions regarding allocations and the like may be better suited for Kristi.

src/coldfront_plugin_cloud/esi.py
QuanMPhm: Multipliers and static additions refer to what an "Allocation Unit" for an ESI allocation represents. For other resources that can be allocated like [Openstack](https://github.com/nerc-project/coldfront-plugin-cloud/blob/6228541b18ca776124d5096ff8ecca2109df1634/src/coldfront_plugin_cloud/tasks.py#L21), 1 allocation unit represented 1 CPU, 4G RAM, 20G Disk Storage, etc. If the user asked for 3 allocation units of Openstack, they would be asking for a quota of 3 CPU, 12G RAM, and so on. This means each Openstack resource (CPU, RAM, etc) has a "multiplier" (1, 4G, ...) that is used to determine how much of each resource is in one allocation unit. Additionally, each Openstack allocation has a ["static" amount of certain resources ](https://github.com/nerc-project/coldfront-plugin-cloud/blob/6228541b18ca776124d5096ff8ecca2109df1634/src/coldfront_plugin_cloud/tasks.py#L45C6-L45C15)that are always included, which are 2 Floating IPs. These are the "static additions".

This is to say, I don't know what one allocation unit of ESI would represent, and would therefore like to know the multipliers and static additions for each of its resources (floating IPs, networks, etc)

tzumainn: Okay, yeah, that's more of a policy question - I wouldn't be the right person to answer that.

README.md
QuanMPhm: @knikolla Your thoughts?

src/coldfront_plugin_cloud/esi.py
QuanMPhm: @knikolla Based on our earlier meeting today, I'm guessing 1 network and 1 floating ip is good?

knikolla: Yes.

README.md
knikolla: I guess same things that apply for OpenStack in terms of parameters, since the way Keystone is configured is the same. Just make a new command called `add_esi_resource`.

ci/devstack_esi.sh
knikolla: @QuanMPhm Don't even do a new Devstack cluster, just reuse the same one that is already installed in the CI and don't worry about Ironic for now.

src/coldfront_plugin_cloud/esi.py
knikolla: Just create a new Kyestone project, users, assigns roles, same as OpenStack. Then set quotas, only neutron quotas on networks and floating ips.

src/coldfront_plugin_cloud/management/commands/add_esi_resource.py
knikolla: List the quotas that are specified in the same way that we do for OpenStack. In this case it's only networks and floating ips.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
knikolla: Same as for OpenStack. Is project created and users added correctly. Is quota set correctly.

QuanMPhm:
@knikolla I have finished the integration code for now. For some unrelated reason the microshift test case is not working. I will update the README once we finish code review.

src/coldfront_plugin_cloud/esi.py
```diff
@@ -0,0 +1,52 @@
...
+QUOTA_KEY_MAPPING_ALL_KEYS = dict()
+for service in QUOTA_KEY_MAPPING.keys():
+    QUOTA_KEY_MAPPING_ALL_KEYS.update(QUOTA_KEY_MAPPING[service]['keys'])
+
+
+class ESIResourceAllocator(OpenStackResourceAllocator):
+
+    resource_type = 'esi'
+
+    def set_quota(self, project_id):
```
knikolla, line 31: I'm assuming you copied this function instead of using the one in the base class because it accesses the `QUOTA_KEY_MAPPING` from the `openstack` module.

If you made `QUOTA_KEY_MAPPING` an attribute of the `OpenStackResourceAllocator` class then you can keep overload it in this subclass and therefore keep this function in the base class without copying it.

knikolla (changes requested):

src/coldfront_plugin_cloud/esi.py
QuanMPhm: I decided to leave `QUOTA_KEY_MAPPING` outside the class since making it a class attribute would require making small changes in many places that reference the `QUOTA_KEY_MAPPING`. Will that be fine?

src/coldfront_plugin_cloud/openstack.py
QuanMPhm: This idea is moot for now since we no longer want to query openstack to see what services are available

src/coldfront_plugin_cloud/esi.py
knikolla: I think those small changes are reasonable, considering they will allow us to avoid repeating the logic in `get_quota` and `set_quota`.

joachimweyl: @knikolla is this conversation resolved?

joachimweyl:
@tzumainn once @knikolla approves the ColdFront side are you ready to approve the ESI side?

tzumainn:
@joachimweyl sure - given that ESI is just OpenStack, and I don't have much expertise with ColdFront, I would assume everything just kind of works.

joachimweyl:
@knikolla do you approve the ColdFront side?

src/coldfront_plugin_cloud/management/commands/add_esi_resource.py
```diff
@@ -0,0 +1,100 @@
+from django.core.management.base import BaseCommand
```
knikolla, line 1: I'd rather just use the `add_openstack_resource` function with a new flag for mentioning this is ESI.

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -89,6 +89,10 @@ class CloudAllocationAttribute:
 QUOTA_REQUESTS_GPU = 'OpenShift Request on GPU Quota'
 QUOTA_PVC = 'OpenShift Persistent Volume Claims Quota'
 
+###########################################################
+# ESI Quota Attributes 
+ESI_FLOATING_IPS = 'ESI Floating IP Quota'
```
knikolla, line 94: Just use the OpenStack attribute.

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -89,6 +89,10 @@ class CloudAllocationAttribute:
...
+ESI_NETWORKS = 'ESI Network Quota'
```
knikolla, line 95: Keep the attribute name as OpenStack specific rather than mentioning ESI. Don't make it automatically be added to OpenStack projects though.

src/coldfront_plugin_cloud/esi.py
```diff
@@ -0,0 +1,33 @@
...
+                attributes.ESI_NETWORKS: 'network'
+            }
+        }
+    }
+
+    QUOTA_KEY_MAPPING_ALL_KEYS = {quota_key: quota_name for k in QUOTA_KEY_MAPPING.values() for quota_key, quota_name in k['keys'].items()}
+
+    resource_type = 'esi'
+
+    def get_quota(self, project_id):
```
knikolla, line 26: In the OpenStack base class create a new method called `get_network_quota` and reuse the same logic in both places.

knikolla (changes requested):

src/coldfront_plugin_cloud/management/commands/add_openstack_resource.py
```diff
@@ -1,3 +1,4 @@
+from sre_constants import OP_IGNORE
```
knikolla, line 1: Is this used and if so, what is this used for?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -66,90 +67,88 @@ def check_institution_specific_code(self, allocation, apply):
 
     def handle(self, *args, **options):
 
-        # Openstack Resources first
-        openstack_resources = Resource.objects.filter(
-            resource_type=ResourceType.objects.get(
-                name='OpenStack'
+        # Deal with Openstack and ESI resources
+        for resource_name in ['OpenStack', 'ESI']:
```
knikolla, line 71: A more accurate variable name would be `resource_type`, since this corresponds to the type of resource rather than its name.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -66,90 +67,88 @@ def check_institution_specific_code(self, allocation, apply):
...
+            openstack_resources = Resource.objects.filter(
+                resource_type=ResourceType.objects.get(
+                    name=resource_name
```
knikolla, line 74: I'm curious if you're able to use something like `name__in` here. The addition of the extra level of indentation required by `for resource in [OpenStack, ESI]` messes up the diff enough to make it hard to parse what has changed.

I think you should try to split out the `for allocation in openstack_allocations` loop from inside this loop by creating an `openstack_allocations` list that has allocations for both ESI and OpenStack.

README.md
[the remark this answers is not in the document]
knikolla: This is still unresolved. Please update or remove.

knikolla (changes requested):
Almost there!

src/coldfront_plugin_cloud/management/commands/add_openstack_resource.py
QuanMPhm: That's a mistake :P

README.md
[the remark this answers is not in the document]
knikolla: This is still unresolved. Please update or remove.

knikolla (approved):

knikolla (approved):

src/coldfront_plugin_cloud/openstack.py
```diff
@@ -94,6 +62,35 @@ def get_session_for_resource(resource):
...
+        },
+        'volume': {
+            'keys': {
+                attributes.QUOTA_VOLUMES: 'volumes',
+                attributes.QUOTA_VOLUMES_GB: 'gigabytes',
+            }
+        },
+    }
+
+    QUOTA_KEY_MAPPING_ALL_KEYS = {quota_key: quota_name for k in QUOTA_KEY_MAPPING.values() for quota_key, quota_name in k['keys'].items()}
```
naved001, line 93: I don't understand why this change is needed when the original code is much more readable than this.

src/coldfront_plugin_cloud/tests/functional/esi/test_allocations.py
```diff
@@ -0,0 +1,128 @@
...
+        self.volume = cinderclient.Client(session=self.session, version=3)
+        self.networking = neutronclient.Client(session=self.session)
+        self.role_member = self.identity.roles.find(name='member')
+
+    def test_new_ESI_allocation(self):
+        user = self.new_user()
+        project = self.new_project(pi=user)
+        allocation = self.new_allocation(project, self.resource, 1)
+        allocator = esi.ESIResourceAllocator(self.resource,
+                                                         allocation)
```
naved001, line 34: formatting?

src/coldfront_plugin_cloud/openstack.py
QuanMPhm: I wanted to turn `QUOTA_KEY_MAPPING` and `QUOTA_KEY_MAPPING_ALL_KEYS` into class attributes because I wanted the `esi.py` module to be able to override it, which would prevent some redundant code down the line. 

As for the formatting of `QUOTA_KEY_MAPPING_ALL_KEYS`, that's a formatting oversight on my part. I could make initialize it in the allocator's `__init__()` function if you'd like, though I thought that would lead to a messier diff. 

For now, I've reformatted it to look more proper.

joachimweyl:
@naved001 is this resolved?

naved001 (approved):

PR #150, opened 2024-03-15
Migrating to using upstream keystone oidc plugin

### LINKED ISSUE #149: Migrating discontinued Keycloak container used by Devstack testing
The current Keycloak container by jboss that we are using to test the Openstack plugin is now discontinued, preventing functional testing of our Openstak integration to fail. The Devstack functional tests are also using an [out-of-date OpenID plugin](https://github.com/nerc-project/coldfront-plugin-cloud/blob/6228541b18ca776124d5096ff8ecca2109df1634/ci/devstack.sh#L16). 

From @knikolla's advice, the solution is to migrate to using Keycloak's upstream OpenID plugin, and to use a new Keycloak container.

QuanMPhm:
Closes #149 I edited the devstack functional test to use upstream Keystone's OIDC plugin. A few files from the [old plugin repo](https://github.com/knikolla/devstack-plugin-oidc) were copied and embedded in devstack.sh since I don't know a better place to put them. I am happy to relocate these files given advice.

it is important to note that sometimes the functional test for Openstack fails because one of the [plugin's setup script ](https://github.com/openstack/keystone/blob/4121cf6cb7c3d5e585df827d72e08296664f7326/devstack/tools/oidc/setup_keycloak_client.py#L29)would fail to access an endpoint in the Keycloak container. I am currently investigating the cause, or at least a solution to this failure.

### DIFF
--- a/ci/devstack.sh
+++ b/ci/devstack.sh
@@ -2,6 +2,7 @@
 # Installs Devstack with the OIDC plugin
 #
 set -xe
+REPO_PATH=$PWD
 
 sudo apt-get update
 # sudo apt-get upgrade -y
@@ -9,15 +10,13 @@ sudo apt-get update
 sudo mkdir -p /opt/stack
 sudo chown "$USER:$USER" /opt/stack
 
-if [[ ! "${CI}" == "true" ]]; then
-    sudo apt-get install docker.io docker-compose -y
-fi
-
-git clone https://github.com/knikolla/devstack-plugin-oidc /opt/stack/devstack-plugin-oidc
-source /opt/stack/devstack-plugin-oidc/tools/config.sh
-
-# Start Keycloak
-cd /opt/stack/devstack-plugin-oidc/tools && sudo docker-compose up -d
+# Install CA into keycloak container and host
+mkdir /opt/stack/data
+cd /opt/stack/data
+openssl req -x509 -nodes -newkey rsa:2048 -keyout key.pem -out cert.pem -sha256 -subj "/CN=$(hostname -I | awk '{print $1}')"
+cat cert.pem key.pem > devstack-cert.pem
+sudo cp devstack-cert.pem /usr/local/share/ca-certificates/devstack-cert.crt
+sudo update-ca-certificates
 
 # Install and start Devstack
 git clone https://github.com/openstack/devstack.git /opt/stack/devstack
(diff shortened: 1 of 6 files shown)


ci/devstack.sh
```diff
@@ -44,17 +42,77 @@ echo "
...
+            }
+        ],
+        "remote": [
+            {
+                "type": "OIDC-preferred_username"
+            }
+        ]
+    }
+]
+' >> mapping.json
```
knikolla, line 82: It's fine if this is a separate file rather than being echoed into a file. Call it `devstack_mapping.json`

ci/devstack.sh
```diff
@@ -44,17 +42,77 @@ echo "
...
+    project_name='federated_project',
+    project_domain_name='federated_domain',
+)
+s = session.Session(auth)
+
+if s.get_token():
+    print('Authentication successful!')
+else:
+    sys.exit('OpenID Authentication failed')
+" >> test_oidc_login.py
```
knikolla, line 115: It's fine if this is it's own file rather than being echoed into a file. Call it `devstack_test_oidc_login.py`.

ci/run_functional_tests_openstack.sh
```diff
@@ -3,7 +3,8 @@
 # Tests expect the resource to be name Devstack
 set -xe
 
-source /opt/stack/devstack-plugin-oidc/tools/config.sh
+HOST_IP=`ip addr show eth0 | grep "inet " | awk '{ print $2 }' | awk -F "/"  '{ print $1 }'`
```
knikolla, line 6: This section is repeated in both file, so carve it out into its own file `devstack_config_ip.sh` and use it from there.

knikolla (changes requested):
Looks pretty good! Some minor comments to separate the scripts into own files within the `ci` directory.

QuanMPhm:
I have moved the files embedded in ci/devstack.sh into their own seperate files, as well as moving the code that sets the HOST_IP variable to its own file.

ci/run_functional_tests_openstack.sh
```diff
@@ -2,8 +2,8 @@
 #
 # Tests expect the resource to be name Devstack
 set -xe
-
-source /opt/stack/devstack-plugin-oidc/tools/config.sh
```
knikolla, line 6: This is fine, but no need to use the `$PWD` variable when you can use relative paths. 

```bash
source ci/devstack-config-ip.sh
```

knikolla (approved):

PR #154, opened 2024-04-10
Added management command to migrate field of sciences

QuanMPhm:
I've moved the PR for field of science migration over here from [`coldfront-plugin-api`](https://github.com/nerc-project/coldfront-plugin-api/pull/38)

---

Closes https://github.com/nerc-project/coldfront-nerc/issues/92, I've added a Django management command to allow the migration of field of sciences.
The command requires that a tab-seperated mapping csv is provided, and can allow the migration to new, or already existing fields of sciences.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/migrate_fields_of_science.py
+++ b/src/coldfront_plugin_cloud/management/commands/migrate_fields_of_science.py
@@ -1,0 +1,82 @@
+import csv
+
+from django.core.management.base import BaseCommand
+from coldfront.core.project.models import Project
+from coldfront.core.field_of_science.models import FieldOfScience
+
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """Migrates Coldfront's list of fields of sciences (FOS), changing what
+    FOS a project can be assigned to, and updating the FOS in all existing projects.
+
+    Requires a csv, tab-seperated, containing two columns, the first containing the list of old FOS,
+    the second containing the new FOS that the old FOS will map onto.
+
+    I.e to map 'Quantum Mechanics' and 'Photonics' to 'Physics', provide this csv:
+        Quantum Mechanics   Physics
+        Photonics   Physics
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "-m",
+            "--mapping",
+            required=True,
+            help="required tab-seperated csv file to provide mapping for migration",
+        )
+
+    def handle(self, *args, **options):
+        mapping_csv = options["mapping"]
+        mapping_dict, new_fos_set = self._load_fos_map(mapping_csv)
+        self._validate_old_fos(mapping_dict)
+        self._create_new_fos(new_fos_set)
+        self._migrate_fos(mapping_dict)
+
+        logger.info("Field of science migration completed!")
+
+    @staticmethod
+    def _load_fos_map(mapping_csv):
+        mapping_dict = dict()
+        new_fos_set = set()
+        with open(mapping_csv, "r") as f:
+            rd = csv.reader(f, delimiter="\t")
+            for row in rd:
+                old_fos, new_fos = row
+                mapping_dict[old_fos] = new_fos
+                new_fos_set.add(new_fos)
+
+        return (mapping_dict, new_fos_set)
+
+    @staticmethod
+    def _validate_old_fos(mapping_dict):
+        for old_fos_name in list(mapping_dict.keys()):
+            if not FieldOfScience.objects.filter(description=old_fos_name):
+                logger.warn(f"Old field of science {old_fos_name} does not exist")
+
+    @staticmethod
+    def _create_new_fos(new_fos_set):
+        for new_fos_name in new_fos_set:
+            FieldOfScience.objects.get_or_create(
+                is_selectable=True,
+                description=new_fos_name, 
+            )
+
+    def _migrate_fos(self, mapping_dict):
+        for project in Project.objects.all():
+            cur_fos_name = project.field_of_science.description
+            if cur_fos_name in mapping_dict.keys():
+                new_fos_name = mapping_dict[cur_fos_name]
+                new_fos = FieldOfScience.objects.get(description=new_fos_name)
+                project.field_of_science = new_fos
+                project.save()
+                logger.info(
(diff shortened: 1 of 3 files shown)


src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
knikolla, line 1: I think field should be plural here, fields of science.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
```diff
@@ -0,0 +1,85 @@
...
+from coldfront.core.field_of_science.models import FieldOfScience
+
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """Migrates Coldfront's list of fields of sciences (fos), changing what
```
knikolla, line 14: If you want to go with FOS, the acronym should be capitalized in the occurrences of this help section or when logged or printed out.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
```diff
@@ -0,0 +1,85 @@
...
+    fos a project can be assigned to, and updating the fos in all existing projects.
+
+    Requires a csv, tab-seperated, containing two columns, the first containing the list of old fos,
```
knikolla, line 17: Why did you choose a tab separated CSV instead of a normal CSV?

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
```diff
@@ -0,0 +1,85 @@
...
+        mapping_dict = dict()
+        new_fos_set = set()
+        with open(mapping_csv, "r") as f:
+            rd = csv.reader(f, delimiter="\t")
+            for row in rd:
+                old_fos, new_fos = row
+                mapping_dict[old_fos] = new_fos
+                new_fos_set.add(new_fos)
+
+        return [mapping_dict, new_fos_set]
```
knikolla, line 53: I find returning two values from a function usually not good for readability and easy of understanding of the code base, but if you really want to return like this, do return a tuple.

knikolla: Honestly, you don't need to return both as the `new_fos` set is the unique list of objects of the dict.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @staticmethod
+    def _validate_old_fos(mapping_dict):
+        for old_fos_name in list(mapping_dict.keys()):
+            if not FieldOfScience.objects.filter(description=old_fos_name):
+                logger.warn(f"Old field of science {old_fos_name} does not exist")
+
+    @staticmethod
+    def _create_new_fos(new_fos_set):
```
knikolla, line 62: You can create the new Fields of Science as you iterate over the list in `_migrate_fos`. Change this function to only create one FOS as provided per argument.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
```diff
@@ -0,0 +1,85 @@
...
+        for new_fos_name in new_fos_set:
+            FieldOfScience.objects.get_or_create(
+                is_selectable=True,  # Assuming we want all new fos to be selectable
```
knikolla, line 65: Remove these 2 comments. The line of code is pretty explanatory of what is being assumed.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
```diff
@@ -0,0 +1,85 @@
...
+                description=new_fos_name,  # Also assuming we don't care about NSF codes
+            )
+
+    @staticmethod
+    def _migrate_fos(mapping_dict):
+        # Migrate all projects to new fos
```
knikolla, line 71: This comment is unnecessary as the for loop explains the exact same thing.

knikolla (changes requested):

knikolla:
@QuanMPhm I've updated the description of your PR to contain the original description. Please make sure the PR contains all the information here when you submit it rather than require looking somewhere else.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
QuanMPhm: Looking at Coldfront's list of [Fields of Science](https://github.com/ubccr/coldfront/blob/main/coldfront/core/field_of_science/management/commands/data/field_of_science_data.csv), some of the current FOSs contains commas in their description, like "Atomic, Molecular, and Optical Physics"

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
knikolla: Looked, and the fields of science CSV in upstream ColdFront also is tab-separated. Works for me to be consistent.

src/coldfront_plugin_cloud/management/commands/migrate_fields_of_science.py
```diff
@@ -0,0 +1,81 @@
...
+            description=description,
+        )
+        return new_fos
+
+    def _migrate_fos(self, mapping_dict):
+        for project in Project.objects.all():
+            cur_fos_name = project.field_of_science.description
+            if cur_fos_name in mapping_dict.keys():
+                new_fos_name = mapping_dict[cur_fos_name]
+                new_fos = self._get_or_create_new_fos(description=new_fos_name)
```
knikolla, line 73: I was wrong, you were right to create all fields of science in a separate pass. In this way if an old field of science has not been selected for any project the new one will not be created.

src/coldfront_plugin_cloud/management/commands/migrate_field_of_science.py
knikolla: Sorry, I was wrong, see new comment.

QuanMPhm: Can't believe I reverted my own feature. Sorry, I should have payed more attention.

knikolla (approved):

Issue #163, opened 2024-05-23
Rename attributes to replace GB and MB with GiB and MiB respectively


Milstein:

knikolla:
Closed by #164 

Attribute change for Swift split out into issue #162

PR #164, opened 2024-05-23
renamed the attributes with GiB and MiB removing GB and MB

Milstein:
Fix for #163

The following Allocation attributes are changed:

- OpenStack Volume Quota --> OpenStack Number of Volumes Quota
- OpenStack Compute RAM Quota --> OpenStack Compute RAM Quota (MiB)
- OpenStack Volume GB Quota --> OpenStack Volume Quota (GiB)
- ~~OpenStack Swift Quota in Gigabytes --> OpenStack Swift Quota (GiB)~~

- OpenShift Limit on RAM Quota (MB) --> OpenShift Limit on RAM Quota (MiB)
- OpenShift Limit on Ephemeral Storage Quota (GB) --> OpenShift Limit on Ephemeral Storage Quota (GiB)
- OpenShift Request on Storage Quota (GB) --> OpenShift Request on Storage Quota (GiB)

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -68,24 +68,24 @@ ALLOCATION_ATTRIBUTES = [
 ###########################################################
 # OpenStack Quota Attributes
 QUOTA_INSTANCES = 'OpenStack Compute Instance Quota'
-QUOTA_RAM = 'OpenStack Compute RAM Quota'
+QUOTA_RAM = 'OpenStack Compute RAM Quota (MiB)'
 QUOTA_VCPU = 'OpenStack Compute vCPU Quota'
 
 QUOTA_VOLUMES = 'OpenStack Volume Quota'
-QUOTA_VOLUMES_GB = 'OpenStack Volume GB Quota'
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/attributes.py
```diff
@@ -68,24 +68,24 @@ class CloudAllocationAttribute:
...
+QUOTA_RAM = 'OpenStack Compute RAM Quota (MiB)'
 QUOTA_VCPU = 'OpenStack Compute vCPU Quota'
 
 QUOTA_VOLUMES = 'OpenStack Volume Quota'
-QUOTA_VOLUMES_GB = 'OpenStack Volume GB Quota'
+QUOTA_VOLUMES_GB = 'OpenStack Volume Quota (GiB)'
 
 QUOTA_FLOATING_IPS = 'OpenStack Floating IP Quota'
 
-QUOTA_OBJECT_GB = 'OpenStack Swift Quota in Gigabytes'
```
knikolla, line 79: Remove the swift attribute change from this PR. It will require more work, especially because the quota values in RGW need to he updated. This is not just a label change.

knikolla (changes requested):

knikolla (approved):

PR #168, opened 2024-06-13
Convert Swift quotas from GB to GiB

### LINKED ISSUE #162: Convert swift quotas from GB to GiB
- Convert[ GB_IN_BYTES](https://github.com/nerc-project/coldfront-plugin-cloud/blob/ae311f876143a78ad822f6398ad37b134283279f/src/coldfront_plugin_cloud/openstack.py#L22C1-L22C12) from `10**9 = 1,000,000,000` to `2**30 = 1,073,741,824`.
- Either create a new command to convert the quotas per the new definition of GiB or update `validate_allocations` to set the new values on the allocations. 

Related https://github.com/nerc-project/operations/issues/587

QuanMPhm:
Closes #162. A one-time command is added which will set the Swift quota of every allocation from GB to GiB. The Swift quota attribute name has been appropriately renamed

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/convert_swift_quota_to_gib.py
+++ b/src/coldfront_plugin_cloud/management/commands/convert_swift_quota_to_gib.py
@@ -1,0 +1,56 @@
+import logging
+import re
+
+from coldfront_plugin_cloud import attributes
+from coldfront_plugin_cloud import openstack
+from coldfront_plugin_cloud import openshift
+from coldfront_plugin_cloud import utils
+from coldfront_plugin_cloud import tasks
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db.models import Q
+from coldfront.core.resource.models import (Resource,
+                                            ResourceType)
+from coldfront.core.allocation.models import (Allocation,
+                                              AllocationStatusChoice,
+                                              AllocationUser)
+from keystoneauth1.exceptions import http
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """One time command to convert all Swift quotas on all Openstack allocations from
+    GB to GiB. I.e a Swift quota of 1000^2 bytes will now be 1024^2 bytes"""
+
+    def handle(self, *args, **options):
+        openstack_resources = Resource.objects.filter(
+            resource_type=ResourceType.objects.get(
+                name='OpenStack'
+            )
+        )
+        openstack_allocations = Allocation.objects.filter(
+            Q(status=AllocationStatusChoice.objects.get(name='Active')) | 
+            Q(status=AllocationStatusChoice.objects.get(name='Expired')),
+            resources__in=openstack_resources,
+        )
+        for allocation in openstack_allocations:
+            if not (swift_quota := allocation.get_attribute(attributes.QUOTA_OBJECT_GB)):
+                continue
+
+            allocation_str = f'{allocation.pk} of project "{allocation.project.title}"'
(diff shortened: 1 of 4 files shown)


src/coldfront_plugin_cloud/openstack.py
```diff
@@ -19,7 +19,7 @@
 logger = logging.getLogger(__name__)
 
 # 1 GB = 1 000 000 000 B = 10^9 B
-GB_IN_BYTES = 1000000000
+GB_IN_BYTES = 2 ** 20
```
joachimweyl, line 22: I believe you have a typo 30 not 20?

QuanMPhm: Oh snap. You're right!

src/coldfront_plugin_cloud/management/commands/register_cloud_attributes.py
```diff
@@ -43,6 +43,9 @@
     ('OpenShift Request on Storage Quota (GB)', {
         'name': 'OpenShift Request on Storage Quota (GiB)'
     }),
+    ('OpenStack Swift Quota in Gigabytes', {
```
naved001, line 46: It's actually called [Gibibytes](https://simple.wikipedia.org/wiki/Gibibyte) when powers of 2 are used.

src/coldfront_plugin_cloud/management/commands/convert_swift_quota_to_gib.py
```diff
@@ -0,0 +1,56 @@
...
+                                              AllocationStatusChoice,
+                                              AllocationUser)
+from keystoneauth1.exceptions import http
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """One time command to convert all Swift quotas on all Openstack allocations from
+    Gb to Gib. I.e a Swift quota of 1000^2 bytes will now be 1024^2 bytes"""
```
naved001, line 24: A small "b" means "bits" while a capital "B" represents bytes. So, this should be "...allocations from GB to GiB" so it's consistent.

naved001 (commented):
dumb question, but I don't understand where are we converting this quota? What it looks like is we read "attributes.QUOTA_OBJECT_GB", save it in `swift_quota`, and then set that back again?

src/coldfront_plugin_cloud/management/commands/register_cloud_attributes.py
QuanMPhm: I am not sure what you're pointing out? This entry is for attribute name migrations, so the old attribute name was 'OpenStack Swift Quota in Gigabytes', and I'm renaming it to 'OpenStack Swift Quota (GiB)' on all current allocations.

QuanMPhm:
@naved001 It does look a bit unclear. Because `GB_IN_BYTES` has been [changed](https://github.com/nerc-project/coldfront-plugin-cloud/pull/168/files#diff-b30af4960260039dedc4d4226a3382ccb69357771c093bf0c458c5c503e92181R22) from 10^9 to 2^30, when all the Openstack allocations' Swift quotas are set again, they will be [multiplied](https://github.com/nerc-project/coldfront-plugin-cloud/blob/6a2d53a301dc9e007105b92d7e51e12e99c77e57/src/coldfront_plugin_cloud/openstack.py#L194) by the correct multiplier.

src/coldfront_plugin_cloud/management/commands/register_cloud_attributes.py
naved001: ah, I was mistaken. thanks @QuanMPhm!

naved001 (approved):
@QuanMPhm thanks for the explanation.

PR #171, opened 2024-07-31
Fixed billing with decreasing allocation change requests

### LINKED ISSUE #170: Storge billing is incorrect when quotas are changed by Coldfront change requests
In the case where a storage quota is changed through change requests, this [line](https://github.com/nerc-project/coldfront-plugin-cloud/blob/f7a4dc2f2d212f720fd201bab87971e6c508703d/src/coldfront_plugin_cloud/utils.py#L127) in the billing code would result in the wrong SU hours being calculated. I will submit a PR shortly introducing test cases to surface this bug.

QuanMPhm:
Closes #170. I have introduced 3 test cases for storage billing with change requests (CR), 2 of which are intended to fail. The 3 test cases are:
1. Testing when there is a CR to decrease the quota
2. A CR to increase the quota
3. Multiple CRs to decrease the quota

The test cases involving decreasing the quota should fail, and demonstrates the bug that I will explain below.

From what I read in the billing code, our intention with CRs is that when an allocation's storage quota decreases, this decreased value becomes effective (for billing purposes) when the CR for it is created, not when the CR is approved (or the time at which the quota actually decreased). Due to this [statement](https://github.com/nerc-project/coldfront-plugin-cloud/blob/f7a4dc2f2d212f720fd201bab87971e6c508703d/src/coldfront_plugin_cloud/utils.py#L127), this will not happen, but will lead to a more complicated undesired behavior.

With the current code, the first time the quota decreases, the decreased value becomes effective at the time it is changed. On subsequent instances when the quota decreases, one of two things could happen:
- If the quota is decreased to the same value again, the decreased quota becomes effective from the previous time the quota changed (i.e If the quota was 10GB on May 1st and 5GB on May 10th, the period from May 1st to May 10th would be billed at 5GB)
- Else, the decreased quota becomes effective at time of change.

### DIFF
--- a/src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
+++ b/src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
@@ -1,4 +1,5 @@
 import datetime
+import unittest
 import pytz
 
 import freezegun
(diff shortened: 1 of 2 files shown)


knikolla:
Great work on catching this and apologies for the error in logic! The PR description does match the issue and the effects.

The below conditional should be flipped to `cr_created_at < unbounded_last_event_time` and changing that makes the tests pass correctly.
https://github.com/nerc-project/coldfront-plugin-cloud/blob/f7a4dc2f2d212f720fd201bab87971e6c508703d/src/coldfront_plugin_cloud/utils.py#L127

 Can you add a new commit to this PR that introduces the fix and removes the `expectedFailure`.

QuanMPhm:
@knikolla I've pushed up the bug fix.

knikolla:
@QuanMPhm Thanks, I'll review. If #173 passes checks, I'll merge it quickly so you can rebase your work on top of that.

knikolla (approved):

src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
```diff
@@ -177,6 +178,155 @@ def test_new_allocation_quota_never_approved(self):
...
+            
+        allocation.refresh_from_db()
+
+        value = utils.calculate_quota_unit_hours(
+            allocation,
+            attributes.QUOTA_LIMITS_EPHEMERAL_STORAGE_GB,
+            pytz.utc.localize(datetime.datetime(2020, 3, 1, 0, 0, 1)),
+            pytz.utc.localize(datetime.datetime(2020, 3, 20, 23, 59, 59))
+        )
+        self.assertEqual(value, 384)
```
naved001, line 267: @QuanMPhm how do we get this value?

15th March - Quota is 2
17th March - Changerequest is created to set it to 4
19th March - Quota is set to 4.

so for 2 day the quota was 2 and for the other 2 days it was 4, so I'd expect it to be like 2*48 + 4*48  = 288.

naved001 (commented):
looks good, just a question because I don't understand the math involved.

src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
knikolla: @naved001 The change request creation time is only considered for a change request that decreases the value. This one increases it, therefore for billing purposes the value 4 is only effective from 19th of March to 20th.

So (4 days * 24 hours * 2 SUs) + (2 days * 24 hours * 4 SUs) = 384

naved001: @knikolla I see, thanks. 

So, for 15th of March until 19th it's 2SUs, and then from 19th until the billing end date of 20th of March it's 4SUs.

naved001 (approved):

PR #174, opened 2024-08-05
Support excluding time intervals during storage invoicing and fixed typo

### LINKED ISSUE #166: Implement support for excluding time intervals during an invoicing period
Example, do not charge for storage during the MGHPCC shutdown.

QuanMPhm:
Closes #166. Added an optional CLI argument which allows passing in a txt file containing excluded time intervals. Each line in the txt file should contain a date range that we want to exclude from billing, formatted as
 "%Y-%m-%d,%Y-%m-%d". An example line could be "2023-02-01,2023-02-03".

The time period excluded from billing starts from the first second of the start date, and ends on the last second of the end date. This means it is possible for the excluded date range to start and end on the same day.

Also fixed a minor typo.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
+++ b/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
@@ -95,6 +95,9 @@ class Command(BaseCommand):
                             default='nerc-invoicing')
         parser.add_argument('--upload-to-s3', default=False, action='store_true',
                           help='Upload generated CSV invoice to S3 storage.')
+        parser.add_argument('--excluded-date-ranges', type=str, 
+                            default=None, nargs='+',
+                            help='List of date ranges excluded from billing')
 
     @staticmethod
     def default_start_argument():
@@ -147,7 +150,8 @@ class Command(BaseCommand):
             time = 0
             for attribute in attrs:
                 time += utils.calculate_quota_unit_hours(
-                    allocation, attribute, options['start'], options['end']
+                    allocation, attribute, options['start'], options['end'],
+                    excluded_intervals_list
                 )
             if time > 0:
                 row = InvoiceRow(
@@ -170,6 +174,13 @@ class Command(BaseCommand):
         logger.info(f'Processing invoices for {options["invoice_month"]}.')
         logger.info(f'Interval {options["start"] - options["end"]}.')
 
+        if options["excluded_date_ranges"]:
+            excluded_intervals_list = utils.load_excluded_intervals(
+                options["excluded_date_ranges"]
+            )
+        else:
+            excluded_intervals_list = None
+            
         openstack_resources = Resource.objects.filter(
             resource_type=ResourceType.objects.get(
                 name='OpenStack'
--- a/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
+++ b/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
@@ -208,7 +208,7 @@ class Command(BaseCommand):
 
             for allocation in openstack_allocations:
                 allocation_str = f'{allocation.pk} of project "{allocation.project.title}"'
-                msg = f'Starting billing for for allocation {allocation_str}.'
+                msg = f'Starting billing for allocation {allocation_str}.'
                 logger.debug(msg)
 
                 process_invoice_row(
@@ -219,7 +219,7 @@ class Command(BaseCommand):
 
             for allocation in openshift_allocations:
                 allocation_str = f'{allocation.pk} of project "{allocation.project.title}"'
-                msg = f'Starting billing for for allocation {allocation_str}.'
+                msg = f'Starting billing for allocation {allocation_str}.'
                 logger.debug(msg)
 
                 process_invoice_row(
--- a/src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
+++ b/src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
@@ -11,6 +11,9 @@ from coldfront_plugin_cloud import utils
 from coldfront.core.allocation import models as allocation_models
 
 
+SECONDS_IN_DAY = 3600 * 24
+
+
 class TestCalculateAllocationQuotaHours(base.TestBase):
     def test_new_allocation_quota(self):
         self.resource = self.new_openshift_resource(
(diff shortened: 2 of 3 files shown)


src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -95,6 +95,8 @@ def add_arguments(self, parser):
                             default='nerc-invoicing')
         parser.add_argument('--upload-to-s3', default=False, action='store_true',
                           help='Upload generated CSV invoice to S3 storage.')
+        parser.add_argument('--excluded-date-range-file', type=str, default=None,
+                            help='File containing list of date ranges excluded from billing')
```
knikolla, line 99: Can you please make this accept the arguments directly from the command line rather than a file?

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,51 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
         last_event_value = int(event.value)
 
     # The value remains the same from the last event until the end.
-    since_last_event = math.ceil((end - last_event_time).total_seconds())
+    since_last_event = get_included_duration(last_event_time, end, exclude_interval_list)
     value_times_seconds += since_last_event * last_event_value
 
     return math.ceil(value_times_seconds / 3600)
+
+def load_exlcluded_intervals(excluded_interval_filepath):
```
knikolla, line 172: typo in excluded

knikolla (changes requested):
Wouldn't a simpler implementation be to calculate the several billable periods and then call `calculate_quota_unit_hours` multiple times adding up the result?

knikolla:
> Wouldn't a simpler implementation be to calculate the several billable periods and then call `calculate_quota_unit_hours` multiple times adding up the result?

Nevermind the above comment. I wrote it and then forgot to delete it.

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,51 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
+    with open(excluded_interval_filepath) as f:
+        for line in f:
+            excluded_intervals_list.append(
+                [datetime.datetime.strptime(t, "%Y-%m-%d") 
+                 for t in line.strip().split(",")]
+            )
+
+    return excluded_intervals_list
+
+def get_included_duration(start: datetime.datetime,
```
knikolla, line 183: I think I have a better and simpler idea on how to write this function. Will let you know later today when I have more time.

src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
QuanMPhm: You want the user to pass in a variable amount of excluded date ranges?

knikolla: yes

src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -95,6 +95,9 @@ def add_arguments(self, parser):
                             default='nerc-invoicing')
         parser.add_argument('--upload-to-s3', default=False, action='store_true',
                           help='Upload generated CSV invoice to S3 storage.')
+        parser.add_argument('--excluded-date-ranges', type=str,
```
knikolla, line 98: I believe the way to provide multiple entries by repeating the argument right? `--excluded-date-ranges 2023-04-01,2023-04-02 --excluded-date-ranges 2023-05-01,2023-04-02`.

In that this should be singular `--exluded-date-range`

src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
```diff
@@ -377,3 +377,134 @@ def test_new_allocation_quota_change_request(self):
...
+    def test_calculate_time_excluded_intervals(self):
+        """Test get_included_duration for correctness"""
+        def get_excluded_interval_datetime_list(excluded_interval_list):
+            return [
+                (datetime.datetime(t1[0], t1[1], t1[2], 0, 0, 0), 
+                datetime.datetime(t2[0], t2[1], t2[2], 23, 59, 59)) 
+                for t1, t2 in excluded_interval_list
+            ]
+        
+        SECONDS_IN_DAY = 3600 * 24
```
knikolla, line 390: According to [PEP8](https://www.python.org/dev/peps/pep-0008/#constants) constants should be defined at the module level.

knikolla (changes requested):
Two minor style comments that can be quickly resolved but otherwise looks good! Thanks!

src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
QuanMPhm: I just tested and the way to pass it is by a space-seperated list of arguments. From the [argparse](https://docs.python.org/3/library/argparse.html#nargs) documentation:
```
parser = argparse.ArgumentParser()

parser.add_argument('--foo', nargs='*')

parser.add_argument('--bar', nargs='*')

parser.add_argument('baz', nargs='*')

parser.parse_args('a b --foo x y --bar 1 2'.split())
Namespace(bar=['1', '2'], baz=['a', 'b'], foo=['x', 'y'])
```

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,57 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
+
+def load_excluded_intervals(excluded_interval_arglist):
+    excluded_intervals_list = list()
+    for interval in excluded_interval_arglist:
+        start, end = interval.strip().split(",")
+        excluded_intervals_list.append(
+            [
+                datetime.datetime.strptime(start, "%Y-%m-%d"),
+                datetime.datetime.strptime(end, "%Y-%m-%d").replace(
+                    hour=23, minute=59, second=59)
```
knikolla, line 181: Apologies, I should have caught this in earlier iterations but it only really hit me yesterday evening.

This is inconsistent with how we are treating dates in the rest of this script.

Running the application with `--start 2024-01-01` and `--end 2024-01-02` will produce a 24 interval (from midnight to midnight), however an `--excluded-date-ranges 2024-01-01,2024-01-02` would be a 48 hour interval. 

Regardless of which is the better way, please treat dates as meaning 00:00 on that date to be consistent.

knikolla (changes requested):

src/coldfront_plugin_cloud/utils.py
QuanMPhm: I see. Sorry for not realizing this design choice as well when implementing the feature.

knikolla (approved):

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,56 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
+    total_interval_duration = (end - start).total_seconds()
+    if not excluded_intervals: 
+        return total_interval_duration
+    
+    for e_interval_start, e_interval_end in excluded_intervals:
+        e_interval_start = _clamp_time(e_interval_start, start, end)
+        e_interval_end = _clamp_time(e_interval_end, start, end)
+        total_interval_duration -= (e_interval_end - e_interval_start).total_seconds()
+
+    return math.ceil(total_interval_duration)
```
naved001, line 205: @knikolla it's possible that `total_interval_duration` could become negative if the user inadvertently supplies multiple overlapping intervals to exclude. That's obviously user error but it'd still be nice to add an assertion that `total_interval_duartion >=0`

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,56 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
-    since_last_event = math.ceil((end - last_event_time).total_seconds())
+    since_last_event = get_included_duration(last_event_time, end, exclude_interval_list)
     value_times_seconds += since_last_event * last_event_value
 
     return math.ceil(value_times_seconds / 3600)
+
+def load_excluded_intervals(excluded_interval_arglist):
+    excluded_intervals_list = list()
+    for interval in excluded_interval_arglist:
+        start, end = interval.strip().split(",")
```
naved001, line 176: @knikolla do we want to add any assertions that end time is after the start time just in case the user makes a mistake entering these values?

naved001 (approved):
looks good, just couple of optional suggestions.

joachimweyl:
@QuanMPhm have you had a chance to review Naved's suggestions?

src/coldfront_plugin_cloud/utils.py
QuanMPhm: @naved001 I have solved a medium Leetcode problem and added a function that will merge overlapping excluded time intervals

src/coldfront_plugin_cloud/utils.py
QuanMPhm: @naved001 I have added an assertion to `load_excluded_intervals`. A corresponding test case has also been added.

QuanMPhm:
@naved001 @knikolla I've added some small logic additions to the PR, so I will wait for everyone's approval again.

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,85 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
+    return excluded_intervals_list
+
+def _clamp_time(time, min_time, max_time):
+    if time < min_time:
+        time = min_time
+    if time > max_time:
+        time = max_time
+    return time
+
+def _merge_date_ranges(intervals_list: list):
```
knikolla, line 195: I appreciate your enthusiasm in writing code to support this, but this then means that accepting overlapping intervals as an argument is a supported feature of the codebase, which we don't want to be the case.

Some explanations as to why:
- The admin running the command will not be incentivized to fix the way they are storing these intervals.
- It may also be the case that some other tools we're using do not implement these merging function, therefore using the same interval on OpenStack billing for example will return an error, but not here. 
- This is a complex function that we'd have to keep supporting moving forward in cases of bugs. 

Please simply check for overlapping intervals or unordered start and end and error in case of malformed input.

knikolla (changes requested):

src/coldfront_plugin_cloud/utils.py
QuanMPhm: @knikolla Sorry for taking a while to ask this: Since removing support for merging intervals means overlapping intervals will result in faulty billing, I should throw an error for overlapping intervals as well?

src/coldfront_plugin_cloud/utils.py
```diff
@@ -149,21 +150,73 @@ def calculate_quota_unit_hours(allocation: Allocation,
...
+        start_dt, end_dt = [datetime.datetime.strptime(i, "%Y-%m-%d") for i in [start, end]]
+        assert end_dt > start_dt, f"Interval end date ({end}) is before start date ({start})!"
+        excluded_intervals_list.append(
+            [
+                datetime.datetime.strptime(start, "%Y-%m-%d"),
+                datetime.datetime.strptime(end, "%Y-%m-%d")
+            ] 
+        )
+
+    excluded_intervals_list.sort(key=interval_sort_key)
```
knikolla, line 197: nitpick: You can write this as a lambda.

knikolla (approved):

PR #178, opened 2024-11-01
Avoid interactive prompts during unit tests

larsks:
Always call `initial_setup` with `-f` option to avoid the "Do you want to
proceed?" interactive prompt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 git+https://github.com/CCI-MOC/nerc-rates@74eb4a7#egg=nerc_rates
 boto3
-coldfront >= 1.0.4
+coldfront >= 1.1.0
 python-cinderclient  # TODO: Set version for OpenStack Clients
 python-keystoneclient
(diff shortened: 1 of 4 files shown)


larsks:
This is failing with:

```
django.core.management.base.CommandError: Error: unrecognized arguments: -f
```

How is it failing in GitHub but not locally? I guess I will need to investigate.

larsks:
It was failing because despite the fact that we were requiring `coldfront>=1.0.4`, we were *only* getting coldfront 1.0.4. I've added a commit that increases the minimum version to 1.1.0 (which is still compatible with `>=1.0.4`, so hopefully that's okay).

knikolla:
1.1.0 is okay (in production we're at 1.1.5)

You did end up linting an entire file with your patch though.

larsks:
> You did end up linting an entire file with your patch though.

Whoops, my bad. Let's fix that.

knikolla (approved):

PR #192, opened 2025-02-05
Fix CI

### LINKED ISSUE #191: Fix CI
CI is currently broken. Fix.

knikolla:
- Switched Microstack functional tests to Ubuntu 22.04
- Added ESI testing to MicroStack.
- Removed Devstack testing workflow. This is now redundant since we
  are not using Swift anymore in production but just RGW, and the
  ESI tests were switched to the Microstack job.
- Removed unused scripts in ci/

Closes #191


naved001 (approved):

ci/run_functional_tests_openstack.sh
QuanMPhm, line 1: Sorry if this is a dumb question, but with `test-py38-functional-devstack.yaml` deleted, where is this file being used?

knikolla: Good question. In a future patch I'm planning on reusing that file so that I can remove [this](https://github.com/nerc-project/coldfront-plugin-cloud/blob/103277ae900e89038443e01eddbbe73978312536/.github/workflows/test-py39-functional-microstack.yaml#L35) block from the workflow.

Didn't do it here since I wanted to introduce just the bare minimum changes to unblock CI and not introduce any unnecessary refactoring.

knikolla: @QuanMPhm if this answered your question please mark this thread as resolved.

PR #193, opened 2025-02-24
Allows setting quotas of Openshift virtualized GPUs

QuanMPhm:
Closes this [issue](https://github.com/nerc-project/operations/issues/820#issue-2669094398), which will allow virtualized GPUs to be requested on Coldfront. 

I have added two new Cloud attributes for the virtualized GPUs.

The `validate_allocations` command has been updated to allow retrospectively adding new quota attributes to old allocations.

Until we finish merging the openshift account manager, I have made another PR there to make sure this feature works for now.

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -88,6 +88,8 @@ QUOTA_LIMITS_MEMORY = 'OpenShift Limit on RAM Quota (MiB)'
 QUOTA_LIMITS_EPHEMERAL_STORAGE_GB = 'OpenShift Limit on Ephemeral Storage Quota (GiB)'
 QUOTA_REQUESTS_STORAGE = 'OpenShift Request on Storage Quota (GiB)'
 QUOTA_REQUESTS_GPU = 'OpenShift Request on GPU Quota'
+QUOTA_REQUESTS_GPU_A100_SXM4 = 'OpenShift Request on GPU A100 SXM4 (Only for VM use)'
+QUOTA_REQUESTS_GPU_V100 = 'OpenShift Request on GPU V100 (Only for VM use)'
 QUOTA_PVC = 'OpenShift Persistent Volume Claims Quota'
 
 
@@ -106,5 +108,7 @@ ALLOCATION_QUOTA_ATTRIBUTES = [
     CloudAllocationAttribute(name=QUOTA_LIMITS_EPHEMERAL_STORAGE_GB),
     CloudAllocationAttribute(name=QUOTA_REQUESTS_STORAGE),
     CloudAllocationAttribute(name=QUOTA_REQUESTS_GPU),
+    CloudAllocationAttribute(name=QUOTA_REQUESTS_GPU_A100_SXM4),
+    CloudAllocationAttribute(name=QUOTA_REQUESTS_GPU_V100),
     CloudAllocationAttribute(name=QUOTA_PVC),
 ]
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -65,6 +65,15 @@ class Command(BaseCommand):
                 )
                 logger.warn(f'Attribute "{attr}" added to allocation {alloc_str}')
 
+    def set_default_quota_on_allocation(self, allocation, allocator, coldfront_attr):
+        uqm = tasks.UNIT_QUOTA_MULTIPLIERS[allocator.resource_type]
+        value = allocation.quantity * uqm.get(coldfront_attr, 0)
+        value += tasks.STATIC_QUOTA[allocator.resource_type].get(coldfront_attr, 0)
+        utils.set_attribute_on_allocation(
+            allocation, coldfront_attr, value
+        )
+        return value
+
     def handle(self, *args, **options):
 
         # Deal with Openstack and ESI resources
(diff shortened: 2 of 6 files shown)


QuanMPhm:
@knikolla @jtriley  The microshift functional test is blocked until CCI-MOC/openshift-acct-mgt#103 is merged.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -253,19 +262,30 @@ def handle(self, *args, **options):
...
-                                logger.warning(
-                                    f"Quota for allocation {project_id} was out of date. Reapplied!"
-                                )
-                            except Exception as e:
-                                logger.error(f'setting openshift quota failed: {e}')
-                                continue
+                            if options["apply"]:
+                                current_value = self.set_default_quota_on_allocation(allocation, allocator, attr.name)
+
+                        if not (current_value == expected_value):
```
naved001, line 276: Why didn't you use `elif` here? I don't follow why you changed to else and then have 2 ifs after that. Aren't those 2 `ifs` mutually exclusive?

Unless I misunderstood something, I think this should work and is more readable.

```python
if expected_value is None and current_value is not None:
    # we already handle this
elif expected_value is None and current_value is None:
    # the new case you added
elif not (current_value == expected_value)
    # we already handle this too
```

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -15,6 +15,8 @@
     attributes.QUOTA_LIMITS_EPHEMERAL_STORAGE_GB: lambda x: {":limits.ephemeral-storage": f"{x}Gi"},
     attributes.QUOTA_REQUESTS_STORAGE: lambda x: {":requests.storage": f"{x}Gi"},
     attributes.QUOTA_REQUESTS_GPU: lambda x: {":requests.nvidia.com/gpu": f"{x}"},
+    attributes.QUOTA_REQUESTS_VM_GPU_A100_SXM4: lambda x: {":requests.nvidia.com/A100_SXM4_40GB": f"{x}"}, # TODO: Confirm these are the correct resource names on Openshift
```
knikolla, line 18: @naved001 would these only limit the quota on VMs or also non-VM pods in that namespace?

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -88,6 +88,8 @@ class CloudAllocationAttribute:
 QUOTA_LIMITS_EPHEMERAL_STORAGE_GB = 'OpenShift Limit on Ephemeral Storage Quota (GiB)'
 QUOTA_REQUESTS_STORAGE = 'OpenShift Request on Storage Quota (GiB)'
 QUOTA_REQUESTS_GPU = 'OpenShift Request on GPU Quota'
+QUOTA_REQUESTS_VM_GPU_A100_SXM4 = 'OpenShift Request on VM GPU A100 SXM4 40GB' # TODO: Confirm display name for these attributes
```
QuanMPhm, line 91: @naved001 @knikolla I forgot to ask. Let me know if these values are fine

src/coldfront_plugin_cloud/openshift.py
QuanMPhm, line 18: @naved001 @knikolla ^

QuanMPhm: @joachimweyl I don't have the knowledge to answer that question

joachimweyl: I meant to type @naved001. thank you for catching that.

naved001: @knikolla It would apply to non-VM pods in that namespace as well. 

To add more to it, the limits are only being applied to pods and not the VirtualMachine resource directly i.e. you could create a VirtualMachine resource that exceeds the limits, but then the actual launcher pod is never created because it exceeds the quota which ultimately means the VM won't run. 

This is what happens when I create a VM that exceed a resource in resourcequota
```
➜  virtualization git:(main) ✗ oc get virtualmachine
NAME                  AGE   STATUS     READY
naved-centos-stream   14s   Starting   False
➜  virtualization git:(main) ✗ oc get virtualmachineinstance
NAME                  AGE   PHASE     IP    NODENAME   READY
naved-centos-stream   15s   Pending                    False
➜  virtualization git:(main) ✗ oc get pods
No resources found in naved-project namespace.
➜  virtualization git:(main) ✗ oc get vmi naved-centos-stream -o yaml | yq -r '.status.conditions[1].message'
failed to create virtual machine pod: pods "virt-launcher-naved-centos-stream-h724w" is forbidden: exceeded quota: naved-project-rq, requested: limits.cpu=4010m, used: limits.cpu=0, limited: limits.cpu=3
```

Also, I don't really know how a "regular" pod can even consume this resource.

knikolla: Then if this applies to non-VM pods, we should remove the "VM" prefix from the attribute name.

naved001: Without the VM prefix, it may cause confusion for folks who want to use A100 GPUs for their pods. It might appear that you need that specific quota (QUOTA_REQUESTS_GPU_A100_SXM4) to get A100_SXM4 GPUs for their pods. 

I don't feel strongly about it though.

src/coldfront_plugin_cloud/attributes.py
naved001: looks okay to me.

src/coldfront_plugin_cloud/openshift.py
naved001: looks okay to me.

QuanMPhm:
@knikolla Should I wait for your approval before merging?

src/coldfront_plugin_cloud/openshift.py
naved001: @knikolla 

> Also, I don't really know how a "regular" pod can even consume this resource.

I decided to test this, and while a regular pod can't consume this resource it can absolutely claim it and make it unavailable for others.

@QuanMPhm So I guess we could remove the "VM" from the name there, but the documentation should make it clear that this resource can only be meaningfully consumed by a VM and is not necessary for a regular pod.

QuanMPhm: @naved001  Should I write an issue in `nerc-docs`?

QuanMPhm:
I have removed the TODO comments, and edited the quota attribute for virt GPUs so it indicates its for VM use only:
`'OpenShift Request on GPU A100 SXM4 (Only for VM use)'`

naved001:
@jtriley Now that it's decided that the plan is to have a different openshift cluster just for virtualization, @knikolla suggested that we should expose this as a different resource type since these attributes only make sense for VMs. Do you have any thoughts on this?

QuanMPhm:
Closing this issue as we've now agreed on creating a new resource class for openshift virtualization, instead of extending the openshift resource.

QuanMPhm:
@naved001 @knikolla I know that this PR is closed, but I still feel like[ the ability to set new allocation attributes](https://github.com/nerc-project/coldfront-plugin-cloud/pull/193/files#diff-d7546b1ee6b195f27260aea67dd61837879b4bfc4ee73ed24003e3973196d280R68-R76) to pre-existing allocations might still come in handy. Could I make a PR to add this feature to `validate_allocations`?

PR #201, opened 2025-04-03
Refactored `validate_allocation` to use `tasks.UNIT_QUOTA_MULTIPLIERS` when determining what attributes an allocation should have

QuanMPhm:
As a small improvement, this will make `tasks.UNIT_QUOTA_MULTIPLIERS` the source of truth for what quota attributes any given allocation should have, making the code a bit easier to manage. This will also somewhat simplify the coding required to add the Openshift Virtualization resource type.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -103,39 +103,38 @@ class Command(BaseCommand):
 
             obj_key = openstack.OpenStackResourceAllocator.QUOTA_KEY_MAPPING['object']['keys'][attributes.QUOTA_OBJECT_GB]
 
-            for attr in attributes.ALLOCATION_QUOTA_ATTRIBUTES:
-                if 'OpenStack' in attr.name:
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -52,6 +52,11 @@ def sync_users(project_id, allocation, allocator, apply):
 
         return failed_validation
 
+    @staticmethod
+    def get_expected_attributes(allocator):
```
knikolla, line 56: This should be in `tasks.py`. That's where UQM is in and used from. Also update [these lines](https://github.com/nerc-project/coldfront-plugin-cloud/blob/fcebc93b8388a8769dbbde9c38e8848fc60b0d08/src/coldfront_plugin_cloud/tasks.py#L85-L86) to use it. This way we're not adding a dependence to a variable and its internal structure here, but instead to a function that is well defined and already in use.

knikolla (changes requested):
Besides my 1 comment, looks good.

knikolla (approved):

PR #202, opened 2025-04-03
Implemented feature to add new attributes to existing Openshift allocations

### LINKED ISSUE #200: Implement the feature to add new quota attributes to pre-existing allocations
Although https://github.com/nerc-project/coldfront-plugin-cloud/pull/193 was no longer needed, we still believe the ability to retrospectively add new quota attributes to pre-existing allocations to be useful. This feature would be nice to have in case we ever introduce new quota attributes in the future.

QuanMPhm:
Closes #200. `validate_allocations` will now check if an Openshift allocation
does not have a quota value set on either the Coldfront or Openshift
side. In this case, it will set the default quota value for the
allocation.

Due to the complexity of Openstack quotas, this feature is only
implemented for Openshift allocations for now.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -52,6 +52,11 @@ class Command(BaseCommand):
 
         return failed_validation
 
+    @staticmethod
+    def get_expected_attributes(allocator):
+        resource_name = allocator.resource_type
+        return list(tasks.UNIT_QUOTA_MULTIPLIERS[resource_name].keys())
+
     def check_institution_specific_code(self, allocation, apply):
         attr = attributes.ALLOCATION_INSTITUTION_SPECIFIC_CODE
         isc = allocation.get_attribute(attr)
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -256,19 +266,31 @@ def handle(self, *args, **options):
...
-                                f"Quota for allocation {project_id} was out of date. Reapplied!"
-                            )
-                        except Exception as e:
-                            logger.error(f'setting openshift quota failed: {e}')
-                            continue
+                        if options["apply"]:
+                            expected_value = self.set_default_quota_on_allocation(allocation, allocator, attr)
+                            logger.warning(f"Added default quota for {attr} to allocation {allocation_str} to {expected_value}")
+
+                    if not (current_value == expected_value):
```
naved001, line 281: why did you have to make it an `if` inside of an else and move it down? why couldn't we have done this?

```python
elif not (current_value == expected_value):
    # do the thing for this condition
elif (current_value is None and expected_value is None):
    # do the new thing of setting default quota
```

naved001 (commented):
looks like a good feature to have. Just a question inline

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
QuanMPhm: Funny enough. It seems you [already asked something similar ](https://github.com/nerc-project/coldfront-plugin-cloud/pull/193#discussion_r1970293108)to this before. 

I take it that this means these if statements are very confusing, and I've added the comment block below. @naved001 Should I make an issue to refactor this in the future?

```
# We just checked the case where the quota value is set in the cluster 
# but not in coldfront. This is the only case the cluster value is the 
# "source of truth" for the quota value
# If the coldfront value is set, it is always the source of truth. 
# But first, we need to check if the quota value is set anywhere at all. 
# TODO (Quan): Refactor these if statements so that we can remove this comment block?
```

naved001: @QuanMPhm LMAO. yes, please add a brief comment to explain this otherwise I will, without fail, ask the same question again in a few months.

naved001 (approved):

knikolla (approved):
Looks good to me.

Some comments that don't need to be fixed now:
- I don't love that the behavior of `validate_allocations` between OpenStack and OpenShift is now different. We should try to factor out the commonalities and implement features like this in an agnostic way. 
- Could the `else` have been replaced with an `elif` to keep the depth of the conditional statements flatter?

QuanMPhm:
> Could the else have been replaced with an elif to keep the depth of the conditional statements flatter?

@knikolla Hilariously enough, @naved001 has [already asked a similar question](https://github.com/nerc-project/coldfront-plugin-cloud/pull/193#discussion_r1970293108) before. I'll try to refactor this code block within this year. I could use `elif`, but it would lead to some repeated code :P

PR #204, opened 2025-04-04
Moved inline CI script into appropriate ci script file

QuanMPhm:


ci/run_functional_tests_openstack.sh
```diff
@@ -4,31 +4,21 @@
...
 export DJANGO_SETTINGS_MODULE="local_settings"
 export FUNCTIONAL_TESTS="True"
-export OS_AUTH_URL="http://localhost:5000"
-
-coverage run --source="." -m django test coldfront_plugin_cloud.tests.functional.openstack
-coverage run --source="." -m django test coldfront_plugin_cloud.tests.functional.esi
-coverage report
+export OS_AUTH_URL="https://localhost:5000"
 
-openstack application credential delete $OPENSTACK_DEVSTACK_APPLICATION_CREDENTIAL_ID
```
knikolla, line 34: You should keep the credential deletion.

knikolla (approved):
One comment, otherwise looks good!

ci/run_functional_tests_openstack.sh
knikolla: Also, have the tests be run through `coverage` so that we continue getting coverage percentage.

knikolla: > You should keep the credential deletion.

@QuanMPhm It needs to probably be `microstack.openstack` like the other commands instead of `openstack`, no?

PR #205, opened 2025-04-05
Allow direct communication to Openshift Quota API

### LINKED ISSUE #187: Allow direct communication to Openshift through all quota-related endpoints
Dependant on #177, as the next step to merge `CCI-MOC/openshift-acct-mgt`, we want all functions in the `OpenShiftResourceAllocator` class that modifies Openshift Quota objects to have direct communication with the Openshift cluster. This includes the following functions:
- `set_quota`
- `get_quota`

QuanMPhm:
Closes #187, this PR consists of two commits. The first one fixes an integration bug and will now require Openshift resources to specify two URLs, one to the account manager, and one to the Openshift API. The second implements the integration to the Openshift API. 

More details in the commit messages.

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -19,6 +19,7 @@ class CloudAllocationAttribute:
 
 
 RESOURCE_AUTH_URL = 'Identity Endpoint URL'
+RESOURCE_API_URL = 'OpenShift API Endpoint URL'
 RESOURCE_IDENTITY_NAME = 'OpenShift Identity Provider Name'
 RESOURCE_ROLE = 'Role for User in Project'
 
@@ -33,6 +34,7 @@ RESOURCE_EULA_URL = "EULA URL"
 
 RESOURCE_ATTRIBUTES = [
     CloudResourceAttribute(name=RESOURCE_AUTH_URL),
+    CloudResourceAttribute(name=RESOURCE_API_URL),
     CloudResourceAttribute(name=RESOURCE_IDENTITY_NAME),
     CloudResourceAttribute(name=RESOURCE_FEDERATION_PROTOCOL),
     CloudResourceAttribute(name=RESOURCE_IDP),
--- a/src/coldfront_plugin_cloud/management/commands/add_openshift_resource.py
+++ b/src/coldfront_plugin_cloud/management/commands/add_openshift_resource.py
@@ -17,6 +17,10 @@ class Command(BaseCommand):
                             help='Name of OpenShift resource')
         parser.add_argument('--auth-url', type=str, required=True,
                             help='URL of the openshift-acct-mgt endpoint')
+        parser.add_argument('--api-url', type=str, required=True,
+                            help='API URL of the openshift cluster')
+        parser.add_argument('--idp', type=str, required=True,
+                            help='Name of Openshift identity provider')
         parser.add_argument('--role', type=str, default='edit',
                             help='Role for user when added to project (default: edit)')
 
@@ -36,6 +40,18 @@ class Command(BaseCommand):
                 name=attributes.RESOURCE_AUTH_URL),
             resource=openshift,
             value=options['auth_url']
+        )
+        ResourceAttribute.objects.get_or_create(
+            resource_attribute_type=ResourceAttributeType.objects.get(
+                name=attributes.RESOURCE_API_URL),
+            resource=openshift,
+            value=options['api_url']
+        )
+        ResourceAttribute.objects.get_or_create(
+            resource_attribute_type=ResourceAttributeType.objects.get(
+                name=attributes.RESOURCE_IDENTITY_NAME),
+            resource=openshift,
+            value=options['idp']
         )
         ResourceAttribute.objects.get_or_create(
             resource_attribute_type=ResourceAttributeType.objects.get(
--- a/src/coldfront_plugin_cloud/openshift.py
+++ b/src/coldfront_plugin_cloud/openshift.py
@@ -77,7 +77,7 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
     def k8_client(self):
         # Load Endpoint URL and Auth token for new k8 client
         openshift_token = os.getenv(f"OPENSHIFT_{self.safe_resource_name}_TOKEN")
-        openshift_url = self.resource.get_attribute(attributes.RESOURCE_AUTH_URL)
+        openshift_url = self.resource.get_attribute(attributes.RESOURCE_API_URL)
 
         k8_config = kubernetes.client.Configuration()
         k8_config.api_key["authorization"] = openshift_token
@@ -181,7 +181,7 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
     def get_federated_user(self, username):
         if (
             self._openshift_user_exists(username)
-            and self._openshift_get_identity(username)
+            and self._openshift_identity_exists(username)
             and self._openshift_useridentitymapping_exists(username, username)
         ):
             return {'username': username}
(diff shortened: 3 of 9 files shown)


src/coldfront_plugin_cloud/openshift.py
```diff
@@ -264,22 +264,33 @@ def _openshift_get_identity(self, id_user):
     def _openshift_user_exists(self, user_name):
         try:
             self._openshift_get_user(user_name)
-        except kexc.NotFoundError:
-            return False
+        except kexc.NotFoundError as e:
+            # Ensures error raise because resource not found, 
+            # not because of other reasons, like incorrect url
+            e_info = json.loads(e.body)
+            if e_info.get("reason") == "NotFound":
```
naved001, line 315: if the URL is invalid then what will `e_info.get("reason")` have?

naved001: since we are trying to make this more robust, does it make sense to check any additional details to confirm that we actually got "NotFound" after we hit the right URL/API/Resource?

like any of these
```
>>> e_info["details"]["name"]
'naved002' 
>>> e_info["details"]["group"]
'user.openshift.io'
>>> e_info["details"]["kind"]
'users'
```

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -146,20 +146,64 @@ def create_project(self, suggested_project_name):
...
+        for quota_name, quota_value in quota_from_project.items():
+            if quota_value:
+                quota[quota_name] = quota_value
+
+        quota_object = {
+            "Version": "0.9",
+            "Kind": "MocQuota",
+            "ProjectName": project_id,
+            "Quota": quota,
+        }
```
naved001, line 205: I don't understand this quota_object. Why does it have a Kind and Version? and why is it version 0.9?

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -297,3 +341,48 @@ def _openshift_useridentitymapping_exists(self, user_name, id_user):
             for identity in user.get("identities", [])
         )
     
+    def _openshift_get_project(self, project_name):
+        api = self.get_resource_api(API_PROJECT, "Project")
+        return clean_openshift_metadata(api.get(name=project_name).to_dict())
+
+    def _openshift_get_resourcequotas(self, project_id):
+        """Returns a list of all of the resourcequota objects"""
```
naved001, line 349: "All resourcequota objects in namespace with `project_id`"

naved001 (commented):
mostly looks good, but just some comments/questions

src/coldfront_plugin_cloud/openshift.py
QuanMPhm: > if the URL is invalid then what will `e_info.get("reason")` have?

If the URL is invalid, as in it wasn't the Openshift API but some other web service (like the account manager), `_openshift_get_user()` would naturally encounter a 404 error. When the Openshift client encounters a 404, it raises a `kexc.NotFoundError`. In the case where a `kexc.NotFoundError ` happened because the URL is invalid, the error object would just contain the 404 page returned by the server that we attempted to communicate with.

In the case of the account manager, printing out `e.body` gave something like this:
```
(b'<!doctype html>\n<html lang=en>\n<title>404 Not Found</title>\n<h1>Not Foun'
 b'd</h1>\n<p>The requested URL was not found on the server. If you entered '
 b'the URL manually please check your spelling and try again.</p>\n')
```



> since we are trying to make this more robust, does it make sense to check any additional details to confirm that we actually got "NotFound" after we hit the right URL/API/Resource?

I'll add these additional checks. They makes sense

src/coldfront_plugin_cloud/openshift.py
QuanMPhm: This quota object is the object that [the account manager returns](https://github.com/CCI-MOC/openshift-acct-mgt/blob/42db8f80962fd355eac1bc80a1894dc6bb824f12/acct_mgt/moc_openshift.py#L335-L340). As for why it was decided that the quota object would have this format, I have no clue. Looking at the original [PR](https://github.com/CCI-MOC/openshift-acct-mgt/pull/41) that introduced this, it's not obvious why this object format was chosen. I just decided to move over all the relevant bits from the account manager with minimum changes.

QuanMPhm:
`_openshift_user_exists()` has been edited to be a bit more robust. Some minor changes to docstrings. Changed the `test_get_federated_user_not_exist` test case to check for expected behavior when a `kexc.NotFoundError` error is actually thrown.

QuanMPhm:
The way the Openshift quota is fetched has been simplified. `_get_moc_quota_from_resourcequotas()` has been removed, and the weird quota object has been removed. Getting the Openshift quota now just returns a dict with quota attributes and values

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -146,20 +146,43 @@ def create_project(self, suggested_project_name):
...
+
+        logger.info(f"Quota for {project_id} successfully created")
 
     def get_quota(self, project_id):
-        url = f"{self.auth_url}/projects/{project_id}/quota"
-        r = self.session.get(url)
-        return self.check_response(r)
+        cloud_quotas = self._openshift_get_resourcequotas(project_id)
+        combined_quota = {}
+        # TODO: {Quan} Do our project namespace have more than one quota object?
```
QuanMPhm, line 180: Answered by @naved001 in https://github.com/nerc-project/coldfront-plugin-cloud/pull/207#discussion_r2039889753

hpdempsey:
What are the circumstances under which this direct communication is needed?  The description is insufficient.

QuanMPhm:
@hpdempsey Currently, this Coldfront plugin interacts with the Openshift API through a piece of middleware, [`openshift-acct-mgt`](https://github.com/cci-moc/openshift-acct-mgt). The motivation for this middleware, and why we're now planning to bypass it, is detailed in @knikolla's comment [here](https://github.com/cci-moc/openshift-acct-mgt). I would have explained it myself but I'm worried of wording it incorrectly. :P

@knikolla @naved001 I'll defer to you guys for more clarification

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -33,6 +34,7 @@ class CloudAllocationAttribute:
 
 RESOURCE_ATTRIBUTES = [
     CloudResourceAttribute(name=RESOURCE_AUTH_URL),
+    CloudResourceAttribute(name=RESOURCE_API_URL),
```
knikolla, line 37: Note to self: when cutting a new release, specify this in Release Notes.
> Already existing OpenShift Resources need to have the `OpenShift API Endpoint URL` and `OpenShift Identity Provider Name` attribute added.

src/coldfront_plugin_cloud/management/commands/add_openshift_resource.py
```diff
@@ -17,6 +17,10 @@ def add_arguments(self, parser):
                             help='Name of OpenShift resource')
         parser.add_argument('--auth-url', type=str, required=True,
                             help='URL of the openshift-acct-mgt endpoint')
+        parser.add_argument('--api-url', type=str, required=True,
+                            help='API URL of the openshift cluster')
```
knikolla, line 21: nit: Don't fix here, we can follow up in a successive patch. Properly capitalize OpenShift in these 2 arguments.

src/coldfront_plugin_cloud/openshift.py
knikolla: Returning the combined Quota as you are doing now seems fair and preserves the current behavior of what `acct-mgt` was doing. Please remove the comments.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -146,20 +146,43 @@ def create_project(self, suggested_project_name):
         project_name = project_id
         self._create_project(project_name, project_id)
         return self.Project(project_name, project_id)
+    
+    def delete_moc_quotas(self, project_id):
```
knikolla, line 150: nit: I know this was called like this in `acct-mgt`, can be renamed in a follow-up. It shouldn't have moc in the name as there's nothing MOC specific there.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -264,25 +287,84 @@ def _openshift_get_identity(self, id_user):
...
         try:
             self._openshift_get_user(user_name)
-        except kexc.NotFoundError:
-            return False
+        except kexc.NotFoundError as e:
+            # Ensures error raise because resource not found, 
+            # not because of other reasons, like incorrect url
+            e_info = json.loads(e.body)
+            if (
+                e_info["reason"] == "NotFound"
```
knikolla, line 295: At least the `e_info["reason"] == "NotFound"` of the following 3 functions can be merged together in a helper function. This is more than a nitpick, but can be done in a follow-up as it's not enough to warrant holding off this patch.

knikolla (approved):

QuanMPhm:
> This looks fine to me. I know you have added some unit tests, but do you think the existing functional tests sufficiently test the resourcequota apis?

@naved001 The unit test `test_quota.py` was added to make sure the allocator's functions returned the expected output. You're right in pointing out that the functional tests would cover the quota API, but I thought it would also be prudent to add unit tests as well.

src/coldfront_plugin_cloud/openshift.py
QuanMPhm: I've made an issue listing all of your nits

src/coldfront_plugin_cloud/management/commands/add_openshift_resource.py
QuanMPhm: @knikolla Just so I'm clear about this comment, you want fix the capitalization of these cli options, and you want this fix to be done in future PR, not here?

Initially, I thought you wanted me to remove this change entirely :P

naved001 (approved):

PR #211, opened 2025-04-21
Allow adding `nerc.mghpcc.org/allow-unencrypted-routes: 'true'` label to all pre-existing and future Openshift projects

### LINKED ISSUE #196: Need to set additional `nerc.mghpcc.org/allow-unencrypted-routes: 'true'` label by default to any newly approved or already existing OpenShift allocations
All OCP namespaces that want to use the `letsencrypt-production-http01` cluster-issuer in their Ingress with a **DNS CNAME** record must include the label `nerc.mghpcc.org/allow-unencrypted-routes: 'true'` otherwise they **will not** be able to use it.

Since we have some live DNS-based Ingress running in production, we need to carefully verify and apply the `nerc.mghpcc.org/allow-unencrypted-routes: 'true'` label only where necessary, ensuring no duplicates while maintaining uninterrupted service.

Related to [this issue](https://github.com/nerc-project/operations/issues/814#issuecomment-2666871468)

QuanMPhm:
Closes #196. Built on top #205. This PR consists on the last commit. More info in the commit message.

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
+++ b/src/coldfront_plugin_cloud/management/commands/validate_allocations.py
@@ -52,6 +52,22 @@ class Command(BaseCommand):
 
         return failed_validation
 
+    @staticmethod
+    def sync_openshift_project_labels(project_id, allocator, apply):
+        cloud_namespace_obj = allocator._openshift_get_namespace(project_id)
+        cloud_namespace_obj_labels = cloud_namespace_obj["metadata"]["labels"]
+        if missing_or_incorrect_labels := [
+            label_items[0] for label_items in openshift.PROJECT_DEFAULT_LABELS.items() if 
+            label_items not in cloud_namespace_obj_labels.items()
+        ]:
+            logger.warning(
+                f"Openshift project {project_id} is missing default labels: {', '.join(missing_or_incorrect_labels)}"
+            )
+            cloud_namespace_obj_labels.update(openshift.PROJECT_DEFAULT_LABELS)
+            if apply:
+                allocator.patch_project(project_id, cloud_namespace_obj)
+                logger.warning(f"Labels updated for Openshift project {project_id}: {', '.join(missing_or_incorrect_labels)}")
+
     def check_institution_specific_code(self, allocation, apply):
         attr = attributes.ALLOCATION_INSTITUTION_SPECIFIC_CODE
         isc = allocation.get_attribute(attr)
@@ -183,6 +199,7 @@ class Command(BaseCommand):
             quota = allocator.get_quota(project_id)
 
             failed_validation = Command.sync_users(project_id, allocation, allocator, options["apply"])
+            Command.sync_openshift_project_labels(project_id, allocator, options["apply"])
 
             for attr in tasks.get_expected_attributes(allocator):
                 key_with_lambda = allocator.QUOTA_KEY_MAPPING.get(attr, None)
--- a/src/coldfront_plugin_cloud/openshift.py
+++ b/src/coldfront_plugin_cloud/openshift.py
@@ -27,6 +27,14 @@ IGNORED_ATTRIBUTES = [
     "uid",
 ]
 
+
+PROJECT_DEFAULT_LABELS = {
+    'opendatahub.io/dashboard': "true",
+    'modelmesh-enabled': "true",
+    'nerc.mghpcc.org/allow-unencrypted-routes': "true"
+}
+
+
 def clean_openshift_metadata(obj):
     if "metadata" in obj:
         for attr in IGNORED_ATTRIBUTES:
@@ -146,6 +154,9 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
         self._create_project(project_name, project_id)
         return self.Project(project_name, project_id)
     
+    def patch_project(self, project_id, new_project_spec):
+        self._openshift_patch_namespace(project_id, new_project_spec)
+    
     def delete_moc_quotas(self, project_id):
         """deletes all resourcequotas from an openshift project"""
         resourcequotas = self._openshift_get_resourcequotas(project_id)
@@ -238,14 +249,10 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
         headers = {"Content-type": "application/json"}
         annotations = {"cf_project_id": str(self.allocation.project_id),
                        "cf_pi": self.allocation.project.pi.username}
-        labels = {
-            'opendatahub.io/dashboard': "true",
-            'modelmesh-enabled': "true",
-        }
 
         payload = {"displayName": project_name,
                    "annotations": annotations,
-                   "labels": labels}
+                   "labels": PROJECT_DEFAULT_LABELS}
         r = self.session.put(url, data=json.dumps(payload), headers=headers)
         self.check_response(r)
 
@@ -323,6 +330,15 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
     def _openshift_get_project(self, project_name):
         api = self.get_resource_api(API_PROJECT, "Project")
         return clean_openshift_metadata(api.get(name=project_name).to_dict())
+    
+    def _openshift_get_namespace(self, namespace_name):
+        api = self.get_resource_api(API_CORE, "Namespace")
+        return clean_openshift_metadata(api.get(name=namespace_name).to_dict())
+
+    def _openshift_patch_namespace(self, project_name, new_project_spec):
+        # During testing, apparently we can't patch Projects, but we can do so with Namespaces
+        api = self.get_resource_api(API_CORE, "Namespace")
+        res = api.patch(name=project_name, body=new_project_spec)
 
     def _openshift_get_resourcequotas(self, project_id):
         """Returns a list of resourcequota objects in namespace with name `project_id`"""
--- a/src/coldfront_plugin_cloud/tests/functional/openshift/test_allocation.py
+++ b/src/coldfront_plugin_cloud/tests/functional/openshift/test_allocation.py
@@ -214,3 +214,35 @@ class TestAllocation(base.TestBase):
         })
 
         allocator._get_role(user.username, project_id)
+
+    def test_project_default_labels(self):
+        user = self.new_user()
+        project = self.new_project(pi=user)
+        allocation = self.new_allocation(project, self.resource, 1)
+        allocator = openshift.OpenShiftResourceAllocator(self.resource,
+                                                         allocation)
+
+        tasks.activate_allocation(allocation.pk)
+        allocation.refresh_from_db()
+
+        project_id = allocation.get_attribute(attributes.ALLOCATION_PROJECT_ID)
+
+        # Check project labels
+        namespace_dict_labels = allocator._openshift_get_namespace(project_id)["metadata"]["labels"]
+        self.assertTrue(namespace_dict_labels.items() > openshift.PROJECT_DEFAULT_LABELS.items())
+
+        # What if we have a new custom label, or changed value?
+        openshift.PROJECT_DEFAULT_LABELS["test"] = "test"
+        call_command('validate_allocations', apply=True)
+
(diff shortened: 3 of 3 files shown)


src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -52,6 +52,16 @@ def sync_users(project_id, allocation, allocator, apply):
 
         return failed_validation
 
+    @staticmethod
+    def sync_openshift_project_labels(project_id, allocator, apply):
+        cloud_namespace_obj = allocator._openshift_get_namespace(project_id)
+        cloud_namespace_obj_labels = cloud_namespace_obj["metadata"]["labels"]
```
knikolla, line 58: Is the `labels` key guaranteed to exist for a project that has no labels?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -52,6 +52,16 @@ def sync_users(project_id, allocation, allocator, apply):
...
+        if not (cloud_namespace_obj_labels.items() >= openshift.PROJECT_DEFAULT_LABELS.items()):
```
knikolla, line 59: why `not >=` instead of `<=`?

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -52,6 +52,16 @@ def sync_users(project_id, allocation, allocator, apply):
...
+            cloud_namespace_obj_labels.update(openshift.PROJECT_DEFAULT_LABELS)
+            if apply:
```
knikolla, line 61: You need to print a message even outside of `apply`. The reason for running the command without `apply` is so that it would notify you of exactly what it would do if you were to apply it. See the other cases below.

For example here (just before this apply) you can print the labels that were missing.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
```diff
@@ -52,6 +52,16 @@ def sync_users(project_id, allocation, allocator, apply):
...
+                allocator._openshift_replace_namespace(project_id, cloud_namespace_obj)
+                logger.warning(f"Labels updated for Openshift project {project_id}")
```
knikolla, line 63: This warning should say which labels were missing.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -147,6 +153,10 @@ def create_project(self, suggested_project_name):
         self._create_project(project_name, project_id)
         return self.Project(project_name, project_id)
     
+    def replace_project(self, project_name, new_project_spec):
+        sanitized_project_name = utils.get_sanitized_project_name(project_name)
```
knikolla, line 157: You need to use the project ID here that was saved as the attribute ALLOCATED_PROJECT_ID, otherwise a change in the `get_sanitized_project_name` function will point you to a different project.

knikolla (changes requested):
Made some comments. Otherwise looks pretty close.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
QuanMPhm: Running some tests with openshift, it seems even if I patch the namespace object such that it doesn't have labels, openshift [automatically](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name) adds a single label to the namespace, `kubernetes.io/metadata.name`

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
QuanMPhm: Because I am silly. I'm sorry :((

src/coldfront_plugin_cloud/openshift.py
QuanMPhm: In `validate_allocations`, the command is using the project ID from `ALLOCATED_PROJECT_ID`. I suppose the argument name `project_name` is a bit confusing. It's derived from the coldfront project name, sanitized, then with a random hex string appended. I've changed the argument name is `project_id` instead.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
naved001: This is actually interesting because this doesn't look like a good way to compare dictionaries.

src/coldfront_plugin_cloud/tests/functional/openshift/test_allocation.py
```diff
@@ -214,3 +214,26 @@ def test_reactivate_allocation(self):
...
+                                                         allocation)
+
+        tasks.activate_allocation(allocation.pk)
+        allocation.refresh_from_db()
+
+        project_id = allocation.get_attribute(attributes.ALLOCATION_PROJECT_ID)
+
+        # Check project labels
+        namespace_dict_labels = allocator._openshift_get_namespace(project_id)["metadata"]["labels"]
+        self.assertTrue(namespace_dict_labels.items() > openshift.PROJECT_DEFAULT_LABELS.items())
```
naved001, line 232: It's really confusing when using > or < to compare dictionary items. I think the recommended approach is to cast to sets and compare those instead that way you are explicit in what's being compared.

naved001: I am going to mark this as resolved, because we are checking if one is a subset of another (so we need to compare both, key and the value) so this is okay.

QuanMPhm: After our slack conversation, I'll wait for further feedback

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
knikolla: > ~Because I am silly. I'm sorry :((~
> 
> Because if the equality was reversed, it would be testing if `PROJECT_DEFAULT_LABELS` is a superset of `cloud_namespace_obj_labels`. 

I thought that's what we wanted to do here, but now I'm not so sure?

Thinking about it a bit more, I believe the safest thing to do here is to have a `union` of both dictionaries, with values in `PROJECT_DEFAULT_LABELS` taking precedence.

This way
- The admin team can add new labels manually, even outside of ColdFront and not have those extra values be removed or cause a namespace to stay out of sync henceforth.
- When a new label is added, it gets added to all projects.
- When a default value is changed, it's propagated to all projects and overwrites the manual changes.

knikolla: > This is actually interesting because this doesn't look like a good way to compare dictionaries.

It's a bit unintuitive, but it's basically comparing subsets (unless I'm wrong, as I haven't usually used this way of comparing dicts).

`a.items() <= b.items()` is true if a is a subset of b and includes all keys and values of b. 

```python
>>> a = {'L1':1, 'L2':2, 'L3':3}
>>> a_subset = {'L1':1, 'L2':2}
>>> a_almost_subset = {'L1':1, 'L2':3}
>>> a_not_subset = {'L4':4}

>>> a_subset.items() < a.items()
True
>>> a_subset.items() > a.items()
False
>>> a_almost_subset.items() < a.items()
False
>>> a_almost_subset.items() > a.items()
False
>>> a.items() == a.copy().items()
True
>>> a.items() <= a.copy().items()
True
>>> a.items() < a.copy().items()
False
```

QuanMPhm: > Thinking about it a bit more, I believe the safest thing to do here is to have a `union` of both dictionaries, with values in `PROJECT_DEFAULT_LABELS` taking precedence.

@knikolla I believe the current revision of my PR is doing this? The validation function first obtains the list of default labels that are either missing or has a non-default value:
```python
if missing_or_incorrect_labels := [
    label_items[0] for label_items in openshift.PROJECT_DEFAULT_LABELS.items() if 
    label_items not in cloud_namespace_obj_labels.items()
]:
```
If any default labels are missing/non-default, the namespace labels are updated using `PROJECT_DEFAULT_LABELS`:
```python
cloud_namespace_obj_labels.update(openshift.PROJECT_DEFAULT_LABELS)
```



> * The admin team can add new labels manually, even outside of ColdFront and not have those extra values be removed or cause a namespace to stay out of sync henceforth.
> 
> * When a new label is added, it gets added to all projects.
> 
> * When a default value is changed, it's propagated to all projects and overwrites the manual changes.

This code should exhibit the features above.

QuanMPhm:
@naved001 Thank you for pointing that out. I've updated the code to use `patch`. I guess it never dawned on me to use that when testing with the openshift client.

src/coldfront_plugin_cloud/management/commands/validate_allocations.py
knikolla: Ah, I see now.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -146,6 +154,10 @@ def create_project(self, suggested_project_name):
         self._create_project(project_name, project_id)
         return self.Project(project_name, project_id)
     
+    def patch_project(self, project_id, new_project_spec):
+        sanitized_project_name = utils.get_sanitized_project_name(project_id)
```
knikolla, line 158: You need to read this value from the `Allocated Project Name` attribute rather than re-derive it using the function.

If a project has already been create and then `get_sanitized_project_name` function changes to produce a different result than when a project was allocated, the sanitized project name may differ from the `Allocated Project Name` by which the project was created and the namespace was named.

knikolla (changes requested):
One last comment from me, the rest looks good.

src/coldfront_plugin_cloud/openshift.py
QuanMPhm: Since for Openshift, the `Allocated Project Name` is the [same](https://github.com/nerc-project/coldfront-plugin-cloud/blob/c9ef7a1ceb5cd3524eaabba38224556cec4dca40/src/coldfront_plugin_cloud/openshift.py#L145) as the Project ID, I've just changed the `patch_project` function to this:

```python
def patch_project(self, project_id, new_project_spec):
        self._openshift_patch_namespace(project_id, new_project_spec)
```

In `validation_allocations`, [much](https://github.com/nerc-project/coldfront-plugin-cloud/blob/c9ef7a1ceb5cd3524eaabba38224556cec4dca40/src/coldfront_plugin_cloud/management/commands/validate_allocations.py#L175-L185) of the code is already referencing the allocated Openshift project using the `Allocated Project ID` attribute. If it's fine by me I'll continue to use it for `sync_openshift_project_labels`, unless you really want me to use `Allocated Project Name` or do some refactoring.

knikolla (approved):

QuanMPhm:
@knikolla I'll let you resolve the comments and merge

knikolla:
@naved001 can you do one final pass?

src/coldfront_plugin_cloud/tests/functional/openshift/test_allocation.py
```diff
@@ -214,3 +214,26 @@ def test_reactivate_allocation(self):
...
+
+        # What if we have a new custom label, or changed value?
+        openshift.PROJECT_DEFAULT_LABELS["test"] = "test"
+        call_command('validate_allocations', apply=True)
```
naved001, line 236: can we add a test to make sure that a label added outside of coldfront isn't overwritten by a call to `validate_allocations`?

naved001 (commented):
looks good to me, but I'd like to see one small test added (inline)

src/coldfront_plugin_cloud/tests/functional/openshift/test_allocation.py
QuanMPhm: I've added a small test case for this.

naved001 (approved):

PR #215, opened 2025-05-08
Remove python 3.9 checks from CI tests 

KelvinLinBU:
Closes https://github.com/nerc-project/coldfront-plugin-cloud/issues/217. Removed because of successful tool migration to python 3.12.

### DIFF
--- a/.github/workflows/test-functional-microshift.yaml
+++ b/.github/workflows/test-functional-microshift.yaml
@@ -13,9 +13,6 @@ env:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        python-version: [3.9, 3.12]
     runs-on: ubuntu-22.04
 
     steps:
@@ -24,7 +21,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.12
 
       - name: Fix for "grup-efi-amd64-signed" missing
         run: |
--- a/.github/workflows/test-functional-microstack.yaml
+++ b/.github/workflows/test-functional-microstack.yaml
@@ -8,9 +8,6 @@ on:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        python-version: [3.9, 3.12]
     runs-on: ubuntu-22.04
 
     steps:
@@ -19,7 +16,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.12
 
       - name: Install ColdFront and plugin
         run: |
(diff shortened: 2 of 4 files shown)


.github/workflows/test-functional-microshift.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
   build:
     strategy:
       matrix:
-        python-version: [3.9, 3.12]
+        python-version: [3.12]
```
QuanMPhm, line 18: Since there's only one python version. No need to specify the matrix strategy. Just specify the python version in the `setup-python` step

setup.cfg
```diff
@@ -18,7 +18,7 @@ classifiers =
 package_dir =
     = src
 packages = find:
-python_requires = >=3.8
+python_requires = >=3.12
```
QuanMPhm, line 21: Technically, `nerc_rates` [only need 3.11](https://github.com/CCI-MOC/nerc-rates/blob/33701ed0dda369ed468f9cf3f85da07b8b987800/pyproject.toml#L11). Change accordingly

.github/workflows/test-unit.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        python-version: [3.9, 3.12]
+        python-version: [3.12]
```
QuanMPhm, line 13: ^

.github/workflows/test-functional-microstack.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        python-version: [3.9, 3.12]
+        python-version: [3.12]
```
QuanMPhm, line 13: ^

QuanMPhm (changes requested):
Aside from minor comments below, your commit message needs to be amended. We're removing testing for 3.9 not (directly) because we want to use the latest version of `nerc_rates`, but because all our other tools (i.e [coldfront-plugin-api](https://github.com/nerc-project/coldfront-plugin-api/pull/48)) has finished migrating to 3.12. Therefore, it is now 'safe' for us to remove testing for 3.9, since we know all our tools should now all work on 3.12

@knikolla I'll wait for a quick pass from you

KelvinLinBU:
@QuanMPhm changes are done

QuanMPhm (changes requested):
Your commit message summary is very long. This can cause reading issues on certain GUIs or terminals. As a general rule, try to keep the line length of your commit messages before 80 characters. You can check the commit messages in this repo as an example.

Aside from that, this looks good. @knikolla Just like a quick pass from you

QuanMPhm (approved):

knikolla (approved):

PR #219, opened 2025-05-20
Allow direct communication to Openshift Projects API

### LINKED ISSUE #185: Allow direct communication to Openshift through all project-related endpoints
Dependant on #177, as the next step to merge `CCI-MOC/openshift-acct-mgt`, we want all functions in the `OpenShiftResourceAllocator` class that make use of the `project.openshift.io` API  to have direct communication with the Openshift cluster. This includes the following functions:
- `create_project`
- `disable_project`
- `_get_project`

QuanMPhm:
Closes #185. More info in the commit message.

### DIFF
--- a/src/coldfront_plugin_cloud/openshift.py
+++ b/src/coldfront_plugin_cloud/openshift.py
@@ -32,7 +32,17 @@ PROJECT_DEFAULT_LABELS = {
     "opendatahub.io/dashboard": "true",
     "modelmesh-enabled": "true",
     "nerc.mghpcc.org/allow-unencrypted-routes": "true",
+    "nerc.mghpcc.org/project": "true",
 }
+
+LIMITRANGE_DEFAULTS = [
+    {
+        "type": "Container",
+        "default": {"cpu": "1", "memory": "4096Mi", "nvidia.com/gpu": "0"},
+        "defaultRequest": {"cpu": "500m", "memory": "2048Mi", "nvidia.com/gpu": "0"},
+        "min": {"cpu": "125m", "memory": "256Mi"},
+    }
+]
 
 
 def clean_openshift_metadata(obj):
@@ -204,15 +214,14 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
         pass
 
     def disable_project(self, project_id):
-        url = f"{self.auth_url}/projects/{project_id}"
-        r = self.session.delete(url)
-        self.check_response(r)
+        self._openshift_delete_project(project_id)
+        logger.info(f"Project {project_id} successfully deleted")
 
     def reactivate_project(self, project_id):
         project_name = self.allocation.get_attribute(attributes.ALLOCATION_PROJECT_NAME)
         try:
             self._create_project(project_name, project_id)
-        except Conflict:
+        except kexc.ConflictError:
             # This is a reactivation of an already active project
             # most likely for a quota update
             pass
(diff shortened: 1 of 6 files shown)


src/coldfront_plugin_cloud/openshift.py
```diff
@@ -257,20 +266,29 @@ def remove_role_from_user(self, username, project_id):
...
+                "annotations": annotations,
+                "labels": PROJECT_DEFAULT_LABELS,
+            },
         }
-        r = self.session.put(url, data=json.dumps(payload), headers=headers)
-        self.check_response(r)
+
+        self._openshift_create_project(
+            project_def
+        )  # TODO (Quan): How should error handling be done here?
```
QuanMPhm, line 288: @knikolla How should error handling be done when interacting with the Openshift API? Should I make an equivalent to [`check_response()`](https://github.com/nerc-project/coldfront-plugin-cloud/blob/d7760191e8c8e95a9bbd2325a7d18b1ff3480cf8/src/coldfront_plugin_cloud/openshift.py#L124-L138)?

knikolla (approved):

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -257,20 +266,27 @@ def remove_role_from_user(self, username, project_id):
...
+                "labels": PROJECT_DEFAULT_LABELS,
+            },
         }
-        r = self.session.put(url, data=json.dumps(payload), headers=headers)
-        self.check_response(r)
+
+        self._openshift_create_project(project_def)
+        self._openshift_create_limits(project_name)
+
+        logger.info(f"Project {project_id} and limit range successfully created")
```
naved001, line 289: Isn't the project created with the name `project_name` and not `project_id`?

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -257,20 +266,27 @@ def remove_role_from_user(self, username, project_id):
...
 
     def _create_project(self, project_name, project_id):
-        url = f"{self.auth_url}/projects/{project_id}"
-        headers = {"Content-type": "application/json"}
+        pi_username = self.allocation.project.pi.username
+
         annotations = {
             "cf_project_id": str(self.allocation.project_id),
-            "cf_pi": self.allocation.project.pi.username,
+            "cf_pi": pi_username,
```
naved001, line 273: why does this function take in `project_id` as an argument if we can already access it with `self.allocation.project_id`? Are they different, if so, can we add a dosctring that explains the arguments.

src/coldfront_plugin_cloud/tests/unit/openshift/base.py
naved001, line 1: Good call factoring this out and then reusing it!

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -349,6 +363,34 @@ def _openshift_get_project(self, project_name):
...
+        project_name: project_name in which to get LimitRange
+        """
+        api = self.get_resource_api(API_CORE, "LimitRange")
+        return clean_openshift_metadata(api.get(namespace=project_name).to_dict())
+
+    def _openshift_create_limits(self, project_name, limits=None):
+        """
+        project_name: project_name in which to create LimitRange
+        limits: dictionary of limits to create, or None for default
+        """
```
naved001, line 385: I really appreciate having a docstring with all the arguments explained even if it feels trivial. I'd like to see it as a standard practice everywhere.

src/coldfront_plugin_cloud/openshift.py
QuanMPhm: @naved001 Diggin around a bit, it seems for `_create_project(self, project_name, project_id)`, both `project_name` and `project_id` [refer to the same thing](https://github.com/nerc-project/coldfront-plugin-cloud/blob/7d5fdd43551785ef0cd5466d2fbb6421c4039cb3/src/coldfront_plugin_cloud/openshift.py#L163-L164): The Coldfront allocation's name **on Openshift**, (i.e "\<Name of CF Project that owns the allocation\>-\<Random hex values\>"). This decision predates my involvement in the plugin, so I'll let you guys explain that's the case.

With the removal of the account manager from this function, as @naved001 said, the `project_id` argument became unused. I guess I decided to leave it in to avoid having to think about the structural problems involved (explained later below).

`self.allocation.project_id` is the Coldfront project's ID, which is a unique integer that Coldfront's backend database generates for the project object (i.e 10), which has nothing to do with Openshift or `project_id`.

@knikolla @naved001 Within `openshift.py`, it seems the `project_id` and `project_name` arguments with the various functions maybe redundant. As @naved001 mentioned, the allocator has access to the allocation it's affecting, which means it can access the CF project's info by itself. This makes certain code outside of `openshift.py` look a bit odd and out of place (i.e [here](https://github.com/nerc-project/coldfront-plugin-cloud/blob/7d5fdd43551785ef0cd5466d2fbb6421c4039cb3/src/coldfront_plugin_cloud/tasks.py#L114-L115), why does `tasks.py` have to check if the project is already created or not? Can't `openshift.py` and `openstack.py` do that themselves?). Overall, I thought the code looked a bit smelly, but I don't have a concrete refactoring solution to propose. Just something to mention.

QuanMPhm: @naved001 I will remove the `project_id` arg from `_create_project` if you want. That will require minimum changes elsewhere.

naved001: @QuanMPhm It looks like we are using project_name as the display name and project_id is the name of the project/namespace. I think it's probably because openstack had different project IDs and project names, and then we wanted to kind of match that with what we do in openshift even though we get to pick the ID in openshift so we just set them to be the same. 

Let's hold off on making any additional changes for now as I don't want to block your PR for that reasons

PR #222, opened 2025-05-30
Added `Cluster Name` column to storage invoice

QuanMPhm:
It is assumed that all Openshift allocations belong to the production Openstack cluster (`stack`), and Openshift allocations to the prod Openshift cluster (`ocp-prod`)


QuanMPhm:
@knikolla Makes sense. For now, I'll use a script to inject the `Cluster Name` column for the storage invoice then.

PR #226, opened 2025-06-11
Added `Cluster Name` column to storage invoice

### LINKED ISSUE #221: Add cluster name to storage invoice

QuanMPhm:
Closes #221. I just need confirmation on the resource names used in production Coldfront

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
+++ b/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
@@ -41,6 +41,7 @@ class InvoiceRow:
     Project_Name: str = ""
     Project_ID: str = ""
     PI: str = ""
+    Cluster_Name: str = ""
     Invoice_Email: str = ""
     Invoice_Address: str = ""
     Institution: str = ""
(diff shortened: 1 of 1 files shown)


src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -21,6 +21,10 @@
 _RATES = None
 
 
+# What are the cluster names we're using in prod?
+RESOURCE_TO_CLUSTER_NAME_MAPPING = {"?": "stack", "??": "ocp-prod"}
```
QuanMPhm, line 25: @knikolla @naved001 What should be in the resource names here?

knikolla: But again, you shouldn't be doing this here, since you're hardcoding the names of our clusters into the plugin, which we don't want. This conversion should happen downstream.

knikolla (changes requested):
See inline comment. Don't make the translation here. Do it downstream.

src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -190,7 +191,7 @@ def upload_to_s3(s3_endpoint, s3_bucket, file_location, invoice_month):
         logger.info(f"Uploaded to {secondary_location}.")
 
     def handle(self, *args, **options):
-        def process_invoice_row(allocation, attrs, su_name, rate):
+        def process_invoice_row(allocation, attrs, su_name, rate, cluster_name):
```
knikolla, line 194: You can access cluster name from the `allocation` object. No need to pass it in as an argument. This also reduces this change to being 2 lines.

knikolla (changes requested):

PR #228, opened 2025-07-01
Fix Cluster Name in invoice

knikolla:
- Add "Cluster Name" in invoice header
- Write Resource Name instead of Resource Type Name

Related #221


QuanMPhm (approved):
My apologies

PR #233, opened 2025-08-11
Removed references to openshift-acct-mgt

QuanMPhm:
Some clean-up. Now that the account manager is no longer used. Do we plan on nuking that repo anytime soon?


naved001:
> Do we plan on nuking that repo anytime soon?

We'll probably archive it.

PR #241, opened 2025-09-02
This lowers the minimum resources that a container must run with.

naved001:
Some containers in the RHOAI application run with 64Mi of memory. The original reason we set these minimum is to prevent people from running containers without any cpu or memory request.


DanNiESh (approved):

joachimweyl:
@DanNiESh do you have any reasons this should not be included in the Sep 9th maintenance?

joachimweyl:
@knikolla my understanding is that this requires a maintenance to merge. Is that correct?

DanNiESh:
I've manually edit the limit range for student namespaces, so no rush to merge this PR now.

PR #249, opened 2025-10-03
Allow pushing user-allocation membership to Keycloak

### LINKED ISSUE #948: Investigate centralizing authorization for NERC users in Keycloak
## Motivation
Currently, authorization permissions are pushed via ColdFront to downstream services, such as OpenStack and OpenShift.

These permissions should also be centralized, so that they can be consumed automatically. That is currently the case only for the `pi` role, which is described as membership to the `pi` group in Keycloak.

Keycloak has no knowledge of which projects or resource allocations a user or PI has access to, therefore that information can't be included in access_tokens and id_tokens for automating consumption by other services, such as observability, storage, or plugin that automatically sync based on it, etc.

## Completion Criteria
Investigate methods for centralizing authorization information in Keycloak. A decision should be made on the groups or role hierarchy and its representation and meaning.

## Description
- [ ] Investigate available methods for representing authorization information in Keycloak and how they work in the context of RBAC (global) and scoped RBAC (project/resource allocation). 
- [ ] Write a design document based on decided method and get approval.

## Completion dates
Desired - 2025-03-14
Required - TBD

QuanMPhm:
Closes https://github.com/nerc-project/operations/issues/948. More details in the commit message
There are still some questions I have below, so this is still a draft for now.

### DIFF
--- a/ci/setup-keycloak.sh
+++ b/ci/setup-keycloak.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+set -xe
+
+sudo docker rm -f keycloak
+
+sudo docker run -d --name keycloak \
+    -e KEYCLOAK_ADMIN=admin \
+    -e KEYCLOAK_ADMIN_PASSWORD=nomoresecret \
+    -p 8080:8080 \
+    -p 8443:8443 \
+    quay.io/keycloak/keycloak:25.0 start-dev
+
+# wait for keycloak to be ready
+until curl -fsS http://localhost:8080/realms/master; do
+    echo "Waiting for Keycloak to be ready..."
+    sleep 5
+done
+
+# Create client and add admin role to client's service account
+ACCESS_TOKEN=$(curl -X POST "http://localhost:8080/realms/master/protocol/openid-connect/token" \
+     -d "username=admin" \
+     -d "password=nomoresecret" \
+     -d "grant_type=password" \
+     -d "client_id=admin-cli" \
+     -d "scope=openid" \
+| jq -r '.access_token')
+
+
+curl -X POST "http://localhost:8080/admin/realms/master/clients" \
+    -H "Authorization: Bearer $ACCESS_TOKEN" \
+    -H "Content-Type: application/json" \
+    -d '{
+        "clientId": "coldfront",
+        "secret": "nomoresecret",
+        "redirectUris": ["http://localhost:8080/*"],
+        "serviceAccountsEnabled": true
+    }'
+
+COLDFRONT_CLIENT_ID=$(curl -X GET "http://localhost:8080/admin/realms/master/clients?clientId=coldfront" \
+    -H "Authorization: Bearer $ACCESS_TOKEN" | jq -r '.[0].id')
+
+
+COLDFRONT_SERVICE_ACCOUNT_ID=$(curl -X GET "http://localhost:8080/admin/realms/master/clients/$COLDFRONT_CLIENT_ID/service-account-user" \
+    -H "Authorization: Bearer $ACCESS_TOKEN" \
+    -H "Content-Type: application/json" \
+| jq -r '.id')
+
+ADMIN_ROLE_ID=$(curl -X GET "http://localhost:8080/admin/realms/master/roles/admin" \
+    -H "Authorization: Bearer $ACCESS_TOKEN" | jq -r '.id')
+
+# Add admin role to the service account user
+curl -X POST "http://localhost:8080/admin/realms/master/users/$COLDFRONT_SERVICE_ACCOUNT_ID/role-mappings/realm" \
+    -H "Authorization: Bearer $ACCESS_TOKEN" \
+    -H "Content-Type: application/json" \
+    -d '[
+            {
+                "id": "'$ADMIN_ROLE_ID'",
+                "name": "admin"
+            }
+        ]'
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -24,7 +24,6 @@ RESOURCE_AUTH_URL = "Identity Endpoint URL"
 RESOURCE_API_URL = "OpenShift API Endpoint URL"
 RESOURCE_IDENTITY_NAME = "OpenShift Identity Provider Name"
 RESOURCE_ROLE = "Role for User in Project"
-RESOURCE_QUOTA_RESOURCES = "Available Quota Resources"
 
 RESOURCE_FEDERATION_PROTOCOL = "OpenStack Federation Protocol"
 RESOURCE_IDP = "OpenStack Identity Provider"
@@ -35,6 +34,8 @@ RESOURCE_DEFAULT_NETWORK_CIDR = "OpenStack Default Network CIDR"
 
 RESOURCE_EULA_URL = "EULA URL"
 RESOURCE_CLUSTER_NAME = "Internal Cluster Name"
+RESOURCE_QUOTA_RESOURCES = "Available Quota Resources"
+RESOURCE_KEYCLOAK_GROUP_TEMPLATE = "Template String for Keycloak Group Names"
 
 RESOURCE_ATTRIBUTES = [
     CloudResourceAttribute(name=RESOURCE_AUTH_URL),
@@ -45,6 +46,7 @@ RESOURCE_ATTRIBUTES = [
     CloudResourceAttribute(name=RESOURCE_PROJECT_DOMAIN),
     CloudResourceAttribute(name=RESOURCE_ROLE),
     CloudResourceAttribute(name=RESOURCE_QUOTA_RESOURCES),
+    CloudResourceAttribute(name=RESOURCE_KEYCLOAK_GROUP_TEMPLATE),
     CloudResourceAttribute(name=RESOURCE_USER_DOMAIN),
     CloudResourceAttribute(name=RESOURCE_EULA_URL),
     CloudResourceAttribute(name=RESOURCE_DEFAULT_PUBLIC_NETWORK),
--- a/src/coldfront_plugin_cloud/base.py
+++ b/src/coldfront_plugin_cloud/base.py
@@ -1,13 +1,17 @@
 import abc
 import functools
 import json
+import logging
 from typing import NamedTuple
 
 from coldfront.core.allocation import models as allocation_models
 from coldfront.core.resource import models as resource_models
 
-from coldfront_plugin_cloud import attributes
+from coldfront_plugin_cloud import attributes, utils
 from coldfront_plugin_cloud.models.quota_models import QuotaSpecs
+
+
+logger = logging.getLogger(__name__)
 
 
 class ResourceAllocator(abc.ABC):
@@ -45,6 +49,94 @@ class ResourceAllocator(abc.ABC):
             user = self.create_federated_user(username)
         return user
 
+    def set_default_quota_on_allocation(self, coldfront_attr):
+        resource_quotaspecs = self.resource_quotaspecs
+        value = resource_quotaspecs.root[coldfront_attr].quota_by_su_quantity(
+            self.allocation.quantity
+        )
+        utils.set_attribute_on_allocation(self.allocation, coldfront_attr, value)
+        return value
+
+    def set_users(self, project_id, apply):
+        coldfront_users = allocation_models.AllocationUser.objects.filter(
+            allocation=self.allocation, status__name="Active"
+        )
+        cluster_users = self.get_users(project_id)
+        failed_validation = False
+
+        # Create users that exist in coldfront but not in the resource
+        for coldfront_user in coldfront_users:
+            coldfront_username = coldfront_user.user.username
+            if coldfront_username not in cluster_users:
+                failed_validation = True
+                logger.info(f"{coldfront_username} is not part of {project_id}")
+                if apply:
+                    self.get_or_create_federated_user(coldfront_username)
+                    self.assign_role_on_user(coldfront_username, project_id)
+
+        # remove users that are in the resource but not in coldfront
+        users = set(
+            [coldfront_user.user.username for coldfront_user in coldfront_users]
+        )
+        for allocation_user in cluster_users:
+            if allocation_user not in users:
+                failed_validation = True
+                logger.info(
+                    f"{allocation_user} exists in the resource {project_id} but not in coldfront"
+                )
+                if apply:
+                    self.remove_role_from_user(allocation_user, project_id)
+
+        return failed_validation
+
+    def check_and_apply_quota_attr(
+        self,
+        attr: str,
+        expected_quota: int | None,
+        current_quota: int | None,
+        apply: bool,
+    ) -> bool:
+        failed_validation = False
+        if current_quota is None and expected_quota is None:
+            msg = (
+                f"Value for quota for {attr} is not set anywhere"
+                f" on {self.allocation_str}"
+            )
+            failed_validation = True
+
+            if apply:
+                expected_quota = self.set_default_quota_on_allocation(attr)
+                msg = f"Added default quota for {attr} to {self.allocation_str} to {expected_quota}"
+            logger.info(msg)
+        elif current_quota is not None and expected_quota is None:
+            msg = (
+                f'Attribute "{attr}" expected on {self.allocation_str} but not set.'
+                f" Current quota is {current_quota}."
+            )
+
+            if apply:
+                utils.set_attribute_on_allocation(self.allocation, attr, current_quota)
+
+                # To pass `current_quota != expected_quota` check
+                expected_quota = current_quota
+
+                msg = f"{msg} Attribute set to match current quota."
+            logger.info(msg)
+
+        if current_quota != expected_quota:
+            msg = (
+                f"Value for quota for {attr} = {current_quota} does not match expected"
+                f" value of {expected_quota} on {self.allocation_str}"
+            )
+            logger.info(msg)
+            failed_validation = True
+
+        return failed_validation
+
+    @functools.cached_property
+    def allocation_str(self):
+        return f'allocation {self.allocation.pk} of project "{self.allocation.project.title}"'
+
     @functools.cached_property
     def auth_url(self):
         return self.resource.get_attribute(attributes.RESOURCE_AUTH_URL).rstrip("/")
@@ -54,7 +146,11 @@ class ResourceAllocator(abc.ABC):
         return self.resource.get_attribute(attributes.RESOURCE_ROLE) or "member"
 
     @abc.abstractmethod
-    def set_project_configuration(self, project_id, dry_run=False):
+    def set_project_configuration(self, project_id, apply=True):
+        pass
+
+    @abc.abstractmethod
+    def get_project(self, project_id):
         pass
 
     @abc.abstractmethod
@@ -83,6 +179,10 @@ class ResourceAllocator(abc.ABC):
 
     @abc.abstractmethod
     def create_federated_user(self, unique_id):
+        pass
+
+    @abc.abstractmethod
+    def get_users(self, unique_id):
         pass
 
     @abc.abstractmethod
--- a/src/coldfront_plugin_cloud/kc_client.py
+++ b/src/coldfront_plugin_cloud/kc_client.py
@@ -1,0 +1,111 @@
+import os
+import functools
+
+import requests
+from pydantic import BaseModel, ConfigDict, RootModel
+
+
+class KeyCloakGroup(BaseModel):
+    """Keycloak group response model"""
+
+    model_config = ConfigDict(extra="allow")
+    id: str
+    name: str
+
+
+class GroupResponse(RootModel):
+    """Wrapper for group list responses"""
+
+    root: list[KeyCloakGroup]
+
+
+class KeyCloakUser(BaseModel):
+    """Keycloak user response model"""
+
+    model_config = ConfigDict(extra="allow")
+    id: str
+    username: str
+
+
+class UserResponse(RootModel):
+    """Wrapper for user list responses"""
+
+    root: list[KeyCloakUser]
+
+
+class KeyCloakAPIClient:
+    def __init__(self):
+        self.base_url = os.getenv("KEYCLOAK_BASE_URL")
+        self.realm = os.getenv("KEYCLOAK_REALM")
+        self.client_id = os.getenv("KEYCLOAK_CLIENT_ID")
+        self.client_secret = os.getenv("KEYCLOAK_CLIENT_SECRET")
+
+        self.token_url = (
+            f"{self.base_url}/realms/{self.realm}/protocol/openid-connect/token"
+        )
+
+    @functools.cached_property
+    def api_client(self):
+        params = {
+            "grant_type": "client_credentials",
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+        }
+        r = requests.post(self.token_url, data=params)
+        r.raise_for_status()
+        headers = {
+            "Authorization": ("Bearer %s" % r.json()["access_token"]),
+            "Content-Type": "application/json",
+        }
+        session = requests.session()
+        session.headers.update(headers)
+        return session
+
+    def create_group(self, group_name):
+        url = f"{self.base_url}/admin/realms/{self.realm}/groups"
+        payload = {"name": group_name}
+        response = self.api_client.post(url, json=payload)
+
+        # If group already exists, ignore and move on
+        if response.status_code not in (201, 409):
+            response.raise_for_status()
+
+    def get_group_id(self, group_name) -> str | None:
+        """Return None if group not found"""
+        query = {
+            "search": group_name,
+            "exact": "true",
+        }
+        url = f"{self.base_url}/admin/realms/{self.realm}/groups"
+        r = self.api_client.get(url, params=query)
+        r.raise_for_status()
+        groups = GroupResponse.model_validate(r.json())
+        return groups.root[0].id if groups.root else None
+
+    def get_user_id(self, cf_username) -> str | None:
+        """Return None if user not found"""
+        # (Quan) Coldfront usernames map to Keycloak usernames
+        # https://github.com/nerc-project/coldfront-plugin-cloud/pull/249#discussion_r2953393852
+        query = {"username": cf_username, "exact": "true"}
+        url = f"{self.base_url}/admin/realms/{self.realm}/users"
+        r = self.api_client.get(url, params=query)
+        r.raise_for_status()
+        users = UserResponse.model_validate(r.json())
+        return users.root[0].id if users.root else None
+
+    def add_user_to_group(self, user_id, group_id):
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups/{group_id}"
+        r = self.api_client.put(url)
+        r.raise_for_status()
+
+    def remove_user_from_group(self, user_id, group_id):
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups/{group_id}"
+        r = self.api_client.delete(url)
+        r.raise_for_status()
+
+    def get_user_groups(self, user_id) -> list[str]:
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups"
+        r = self.api_client.get(url)
+        r.raise_for_status()
+        groups = GroupResponse.model_validate(r.json())
+        return [group.name for group in groups.root]
--- a/src/coldfront_plugin_cloud/openshift.py
+++ b/src/coldfront_plugin_cloud/openshift.py
@@ -206,9 +206,15 @@ class OpenShiftResourceAllocator(base.ResourceAllocator):
         )
         return api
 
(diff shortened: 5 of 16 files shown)


src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,81 @@
...
+    def get_group_id(self, group_name) -> str | None:
+        """Return None if group not found"""
+        query = f"search={group_name}&exact=true"
+        url = f"{self.base_url}/admin/realms/{self.realm}/groups?{query}"
+        r = self.api_client.get(url).json()
+        return r[0]["id"] if r else None
+
+    def get_user_id(self, cf_username) -> str | None:
+        """Return None if user not found"""
+        # TODO (Quan): Confirm that Coldfront usernames map to Keycloak emails, not email, or something else?
```
QuanMPhm, line 66: @knikolla @naved001

QuanMPhm:
@knikolla Two more questions:
1. Do we also want `validate_allocations` to add PIs to pre-existing allocations? 
2. When a PI adds a user to an Coldfront project or allocation, do those users also get added to a the project's Keycloak group?

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,81 @@
...
+        self.client_id = os.getenv("KEYCLOAK_CLIENT_ID", "admin-cli")
+
+        self.token_url = (
+            f"{self.base_url}/realms/{self.realm}/protocol/openid-connect/token"
+        )
+
+    @functools.cached_property
+    def api_client(self):
+        params = {
+            "grant_type": "password",
```
knikolla: See if you can use the client credentials flow instead of admin password here.

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -154,6 +155,17 @@ def set_quota_attributes():
 
         allocator.set_quota(project_id)
 
+        # After setting everything on cluster, add user to Keycloak group
```
knikolla: This function is growing too long (doing too many things) and I think there might a better place for the code in this block.

Perhaps the base class in `base.py` can abstract some of the duties, among `create_project`, `get_or_create_federated_user` and `assign_role_on_user`.

knikolla (changes requested):

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -364,6 +364,8 @@ def assign_role_on_user(self, username, project_id):
             # Role already exists, ignore
             pass
 
+        super().assign_role_on_user(username, project_id)
```
knikolla: Instead of the resource allocators, I think this should be handled in https://github.com/nerc-project/coldfront-plugin-cloud/blob/0bfb50570f6e6c25f685e198cf7a52216b706f57/src/coldfront_plugin_cloud/tasks.py#L89

src/coldfront_plugin_cloud/base.py
```diff
@@ -45,6 +48,29 @@ def get_or_create_federated_user(self, username):
             user = self.create_federated_user(username)
         return user
 
+    def assign_role_on_user(self, username, project_id):
+        self.kc_admin_client.create_group(project_id)
```
knikolla: Instead of using the group ID, how about introducing a new Resource Attribute to Resources that accepts a format string. For example defaulting to `"{resource_name}/{project_name}"`

This would allow operator to specify their own format for group names for the clusters.

So you'd read the resource attribute and then call format on the string providing a few documented options for the available variables.

```python
>>> template = "{resource_name}/{project_name}"
>>> group = template.format(**{"resource_name": "NERC-OCP", "project_name": "1234"}
... )
>>> 
>>> group
'NERC-OCP/1234'
```

knikolla (changes requested):
Did a quick first pass and provided some comments.

Also this needs to be possible configurable via a setting.

knikolla:
@QuanMPhm Actually, another thought, do you think it would make sense to implement this in the Keycloak plugin? https://github.com/nerc-project/coldfront-plugin-keycloak

It could listen to signals in the same way that the cloud plugin listens to signals. It already has a keycloak client implemented. 

And there is nothing in pushing users to a Keycloak group that is specific to either OpenShift or OpenStack.

QuanMPhm:
@knikolla I see that it does make sense to seperate the Keycloak functionality from the rest of the plugin. It makes sense to me. I forgot that repo existed. There would need to be some overhaul to add integration and unit tests to `coldfront-plugin-keycloak`. Is that fine?

knikolla:
> @knikolla I see that it does make sense to seperate the Keycloak functionality from the rest of the plugin. It makes sense to me. I forgot that repo existed. There would need to be some overhaul to add integration and unit tests to `coldfront-plugin-keycloak`. Is that fine?

For now let's keep it here (as not to frontload the work) and we can easily split it out later if needed. Perhaps try implementing it here via signals so as to keep it loosely coupled so that if we need to split it later it doesn't require a lot of uncoupling.

QuanMPhm:
@knikolla @larsks I have a question about the Coldfront Keycloak PR. I know we previously decided to represent user-membership to an allocation by adding them to a Keycloak group with the same name as the `ALLOCATION_PROJECT_ID` . Would it make more sense for the group name to be the allocation's ID on Coldfront instead? This has a stronger guarantee of uniqueness, and we won't have to disambiguate for projects with the same name in different clusters. Developers or tools can correlate the allocation id to the on-cluster project id by using the Coldfront API.

knikolla:
> @knikolla @larsks I have a question about the Coldfront Keycloak PR. I know we previously decided to represent user-membership to an allocation by adding them to a Keycloak group with the same name as the `ALLOCATION_PROJECT_ID` . Would it make more sense for the group name to be the allocation's ID on Coldfront instead? This has a stronger guarantee of uniqueness, and we won't have to disambiguate for projects with the same name in different clusters. Developers or tools can correlate the allocation id to the on-cluster project id by using the Coldfront API.

See my comment here https://github.com/nerc-project/coldfront-plugin-cloud/pull/249#discussion_r2977091803

src/coldfront_plugin_cloud/base.py
QuanMPhm: Apologies. Missed this

QuanMPhm:
@knikolla @larsks I have revised the PR and made the Keycloak integration a toggle-able feature. Calls to the Keycloak API is now triggered through signals. I have questions below that I think would be of interest to you both

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +136,71 @@ def remove_user_from_allocation(allocation_user_pk):
...
+def _get_keycloak_group_name(allocation: Allocation, template_string: str) -> str:
+    """
+    Acceptable variables for the group name template string is:
+    - $resource_name: the name of the resource (e.g. "OpenShift")
+    - $project_name: the name of the project (e.g. "Test Project")
+    """
+    project_name = allocation.project.title
+    resource_name = allocation.resources.first().name
+
+    # TODO (Quan) Do we also want some validation steps for the format string at creation time to make sure its valid?
```
QuanMPhm: @knikolla @larsks And are we fine with the security of `string.Template`?

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +136,71 @@ def remove_user_from_allocation(allocation_user_pk):
...
+    # Do we also want to include allocation id or something at allocation level?
```
QuanMPhm: @knikolla @larsks Right now, I just included resource name and Coldfront project name as allowed variables in the template

src/coldfront_plugin_cloud/signals.py
```diff
@@ -51,8 +57,14 @@ def activate_allocation_user_receiver(sender, **kwargs):
     else:
         add_user_to_allocation(allocation_user_pk)
 
+    if is_keycloak_enabled():
+        add_user_to_keycloak(allocation_user_pk)
```
knikolla: Why are you not executing these asynchronously?

src/coldfront_plugin_cloud/tasks.py
knikolla: The project name is not the right choice as a Project can have several allocations. This should be the `Allocated Project Name` or `Allocated Project ID`. You could potentially pass to the templater all the attributes associated with an allocation.

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,85 @@
...
+    def create_group(self, group_name):
+        url = f"{self.base_url}/admin/realms/{self.realm}/groups"
+        payload = {"name": group_name}
+        response = self.api_client.post(url, json=payload)
+
+        # If group already exists, ignore and move on
+        if response.status_code not in (201, 409):
+            response.raise_for_status()
+
+    def create_user(self, cf_username):
```
knikolla: I would move these helper functions in the test rather than place them here.

knikolla (changes requested):
Quick first pass for direction feedback.

src/coldfront_plugin_cloud/signals.py
QuanMPhm: I forgot to ask :P

knikolla: It was safe to assume considering the structure of the function and the overall symmetry.

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -45,6 +46,7 @@ class CloudAllocationAttribute:
     CloudResourceAttribute(name=RESOURCE_PROJECT_DOMAIN),
     CloudResourceAttribute(name=RESOURCE_ROLE),
     CloudResourceAttribute(name=RESOURCE_QUOTA_RESOURCES),
+    CloudResourceAttribute(name=RESOURCE_KEYSTONE_GROUP_TEMPLATE),
```
knikolla: Keycloak :)

knikolla (changes requested):
Looking much better. Almost there.

A follow-up to this (after we merge the `validate_allocations` refactor) should investigate how to cleanly implement this in `validate_allocations` too.

knikolla:
> Looking much better. Almost there.
> 
> A follow-up to this (after we merge the `validate_allocations` refactor) should investigate how to cleanly implement this in `validate_allocations` too.

Actually... I would make it a whole separate CLI command. `validate_keycloak_groups`.

QuanMPhm:
@knikolla I've allowed the group name template string to accept any allocation attribute. Further documentation is in the docstring for the function `_get_keycloak_group_name`

knikolla:
@naved001 would appreciate a pass from you.

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +140,79 @@ def remove_user_from_allocation(allocation_user_pk):
...
+
+    if (user_id := kc_admin_client.get_user_id(username)) is None:
+        logger.warning(f"User {username} not found in Keycloak, cannot add to group.")
+        return
+
+    group_name_template = allocation.resources.first().get_attribute(
+        attributes.RESOURCE_KEYCLOAK_GROUP_TEMPLATE
+    )
+    if group_name_template is None:
+        logger.warning(
```
knikolla: I had some more time to think about this. Instead of logging a warning, log an `info` message here, this way the presence of the Keycloak Group Template Resource Attribute can act as a flag for enabling the feature for each specific Resource individually. 

Therefore also switch the order so that the group template is parsed before you look up the user.

> Keycloak enabled but no group name template specified for resource <>. Skipping addition to Keycloak group.

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +140,79 @@ def remove_user_from_allocation(allocation_user_pk):
...
```
knikolla, line 211: Same comment as from addition.

knikolla (changes requested):
Had some more time to think about how to expose enabling/disable the feature.

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+
+import requests
+
+
+class KeyCloakAPIClient:
+    def __init__(self):
+        self.base_url = os.getenv("KEYCLOAK_BASE_URL")
+        self.realm = os.getenv("KEYCLOAK_REALM")
+        self.client_id = os.getenv("KEYCLOAK_CLIENT_ID", "coldfront")
+        self.client_secret = os.getenv("KEYCLOAK_CLIENT_SECRET", "nomoresecret")
```
naved001: I'd rather we don't set the default id and secret.

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+    @functools.cached_property
+    def api_client(self):
+        params = {
+            "grant_type": "client_credentials",
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+        }
+        r = requests.post(self.token_url, data=params).json()
```
naved001: maybe some error handling would be useful before you try to access `r["access_token"]`.

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+    def remove_user_from_group(self, user_id, group_id):
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups/{group_id}"
+        r = self.api_client.delete(url)
+        r.raise_for_status()
+
+    def get_user_groups(self, user_id) -> list[str]:
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups"
+        r = self.api_client.get(url)
+        r.raise_for_status()
+        return [group["name"] for group in r.json()]
```
naved001: this may be my limited knowledge of keycloak, but are we guaranteed to have a user be part of some group?

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +140,79 @@ def remove_user_from_allocation(allocation_user_pk):
...
+def _clean_template_string(template_string: str) -> str:
+    return template_string.replace(" ", "_").lower()
+
+
+def _get_keycloak_group_name(allocation: Allocation, template_string: str) -> str:
+    """
+    Acceptable variables for the group name template string is:
+    - $resource_name: the name of the resource (e.g. "OpenShift")
+    - Any allocation attribute defined for the allocation, with spaces replaced by underscores and
+    all lowercase (e.g. for `Project Name`, the variable would be `$project_name`)
```
naved001, line 152: why is there a `$` before `project_name`?

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+            "Content-Type": "application/json",
+        }
+        session = requests.session()
+        session.headers.update(headers)
+        return session
+
+    def create_group(self, group_name):
+        url = f"{self.base_url}/admin/realms/{self.realm}/groups"
+        payload = {"name": group_name}
+        response = self.api_client.post(url, json=payload)
```
naved001, line 67: does this return a group id when a group is created? If yes, we should return that to the caller and we won't need an additional call to get_group_id

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+            "client_secret": self.client_secret,
+        }
+        r = requests.post(self.token_url, data=params).json()
+        headers = {
+            "Authorization": ("Bearer %s" % r["access_token"]),
+            "Content-Type": "application/json",
+        }
+        session = requests.session()
+        session.headers.update(headers)
+        return session
```
naved001, line 62: We are creating and caching this session with a token (and cache the instantiation of the class again later). When does the keycloak token expire? Just wondering how long we can continue to use this. Or is that not a concern since we'll run this thing as needed and then when its needed we start from scratch . I admit I don't know how coldfront tasks call these.

src/coldfront_plugin_cloud/kc_client.py
QuanMPhm: I can't say what the production environment will look like, but local testing doesn't suggest users are automatically added to some group. Regardless, I don't think that should be a problem for consumers of the Keycloak API? @knikolla ?

src/coldfront_plugin_cloud/tasks.py
QuanMPhm: That is the syntax for [Python string templates](https://docs.python.org/3/library/string.html#template-strings-strings), `$` marks the start of placeholders that will be substituted

src/coldfront_plugin_cloud/kc_client.py
QuanMPhm: From local testing and [the online documentation](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_post_adminrealmsrealmgroups), it just returns the status code

src/coldfront_plugin_cloud/kc_client.py
QuanMPhm: used `raise_for_status()`

src/coldfront_plugin_cloud/kc_client.py
QuanMPhm: @naved001 The keycloak API will be called whenever Coldfront's [`remove/activate user`](https://github.com/nerc-project/coldfront-plugin-cloud/blob/0bfb50570f6e6c25f685e198cf7a52216b706f57/src/coldfront_plugin_cloud/signals.py#L46-L58) signals are triggered, which happens when a user is add/removed from an allocation. That being said, I would assume the api_client will be called somewhat often, especially during batch user uploads. 

@naved001 @knikolla Do we still want to cache the api client? Or make a follow-up PR to set a TTL mechanism?

knikolla: Just remove the caching for now.

src/coldfront_plugin_cloud/kc_client.py
knikolla: It is likely that the user may be part of other groups besides the ones corresponding to respective allocations. This is at least guaranteed for PIs who are part of a group called `pi`.

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +140,79 @@ def remove_user_from_allocation(allocation_user_pk):
...
+        allocation.allocationattribute_set.all()
+    )
+
+    template_sub_dict = {"resource_name": resource_name}
+    for attr in allocation_attrs_list:
+        template_sub_dict[
+            _clean_template_string(attr.allocation_attribute_type.name)
+        ] = attr.value
+
+    return Template(template_string).substitute(**template_sub_dict)
```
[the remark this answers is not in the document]
QuanMPhm: @knikolla @naved001 Not raising the error could lead to users being silently added to incorrectly-named Keycloak groups. Is that acceptable?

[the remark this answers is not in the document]
knikolla: The task should break rather than do the wrong thing.

src/coldfront_plugin_cloud/signals.py
```diff
@@ -25,6 +27,10 @@ def is_async():
     return os.getenv("REDIS_HOST")
 
 
+def is_keycloak_enabled():
+    return os.getenv("KEYCLOAK_BASE_URL")
```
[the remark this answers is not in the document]
QuanMPhm: I think one env var is enough to indicate Keycloak is enabled. It's the operator's problem if the config is misconfigured

src/coldfront_plugin_cloud/tasks.py
```diff
@@ -128,3 +140,79 @@ def remove_user_from_allocation(allocation_user_pk):
...
+            f"Keycloak enabled but no group name template specified for resource {allocation.resources.first().name}. Skipping removal from Keycloak group"
+        )
+        return
+
+    if (user_id := kc_admin_client.get_user_id(username)) is None:
+        logger.warning(f"User {username} not found in Keycloak, cannot add to group.")
+        return
+
+    group_name = _get_keycloak_group_name(allocation, group_name_template)
+    group_id = kc_admin_client.get_group_id(group_name)
```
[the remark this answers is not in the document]
QuanMPhm: @knikolla @naved001 This scenario, where the group_id turns out to be `None` will probably happen quite a lot before we add a validation command to add all current allocation users to Keycloak. I'll add the suggestion for now

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+        # https://github.com/nerc-project/coldfront-plugin-cloud/pull/249#discussion_r2953393852
+        query = f"username={cf_username}&exact=true"
+        url = f"{self.base_url}/admin/realms/{self.realm}/users?{query}"
+        r = self.api_client.get(url).json()
+        return r[0]["id"] if r else None
+
+    def add_user_to_group(self, user_id, group_id):
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups/{group_id}"
+        r = self.api_client.put(url)
+        r.raise_for_status()
```
knikolla, line 99: If the user is already part of the group don't error out here, just log an info.

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def remove_user_from_group(self, user_id, group_id):
+        url = f"{self.base_url}/admin/realms/{self.realm}/users/{user_id}/groups/{group_id}"
+        r = self.api_client.delete(url)
+        r.raise_for_status()
```
knikolla, line 104: If the user was not part of the group, don't error out here, just log an info.

knikolla (changes requested):

src/coldfront_plugin_cloud/tasks.py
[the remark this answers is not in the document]
knikolla: Since we already have a lot of allocations and users of those, the validation command is still high priority otherwise only the new group memberships will be recorded and none of the previous ones.

src/coldfront_plugin_cloud/kc_client.py
QuanMPhm: From local testing, it seems if the user is already not part of the group, `204` is returned, so not error is raised.

src/coldfront_plugin_cloud/kc_client.py
QuanMPhm: From local testing, if the user is already part of the group, 204 is returned, so not error is raised.

QuanMPhm:
@knikolla @naved001 I have responded to all comments

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,73 @@
...
+        r = requests.post(self.token_url, data=params)
+        r.raise_for_status()
+        headers = {
+            "Authorization": ("Bearer %s" % r.json()["access_token"]),
+            "Content-Type": "application/json",
+        }
+        session = requests.session()
+        session.headers.update(headers)
+        return session
+
```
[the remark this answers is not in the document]
larsks: @QuanMPhm This seems like a reasonable critique. Looking at the code in `src/coldfront_plugin_cloud/tasks.py`, it looks like there are many operations that will end up requesting a new api client multiple times. I would cache the result of `KeyCloakAPIClient.api_client`...

```python
    @functools.cached_property
    def get_api_client(self):
        params = {
            "grant_type": "client_credentials",
            "client_id": self.client_id,
            "client_secret": self.client_secret,
        }
        r = requests.post(self.token_url, data=params)
        r.raise_for_status()
        headers = {
            "Authorization": ("Bearer %s" % r.json()["access_token"]),
            "Content-Type": "application/json",
        }
        session = requests.session()
        session.headers.update(headers)
        return session
```

...and *remove* the cache from `tasks.get_kc_client`, so that (a) you're only calling `api_client` once per instance of `KeyCloakAPIClient`, and (b) every method in `tasks.py` will start with a fresh api client (because of the call to `get_kc_client`).

At that point, there's no need to `get_kc_client` so you can just write:

```python
    kc_admin_client = kc_client.KeyCloakAPIClient()
```

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,77 @@
...
+        # If group already exists, ignore and move on
+        if response.status_code not in (201, 409):
+            response.raise_for_status()
+
+    def get_group_id(self, group_name) -> str | None:
+        """Return None if group not found"""
+        query = f"search={group_name}&exact=true"
+        url = f"{self.base_url}/admin/realms/{self.realm}/groups?{query}"
+        r = self.api_client.get(url)
```
larsks: Make python build urls for you:

```suggestion
        query = {
          "search": group_name,
          "exact": "true",
        }
        url = f"{self.base_url}/admin/realms/{self.realm}/groups"
        r = self.api_client.get(url, params=query)
```

src/coldfront_plugin_cloud/signals.py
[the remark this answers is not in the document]
larsks: If all of those variables are required we should be validating that they are set and emitting a useful error message if not. On the other hand, this might be the wrong place for it.

Maybe in `KeyCloakAPIClient`, where you read in the variables to set up the client:

```python
class KeyCloakAPIClient:
    def __init__(self):
        self.base_url = os.getenv("KEYCLOAK_BASE_URL")
        self.realm = os.getenv("KEYCLOAK_REALM")
        self.client_id = os.getenv("KEYCLOAK_CLIENT_ID")
        self.client_secret = os.getenv("KEYCLOAK_CLIENT_SECRET")

```

...but have you considered [pydantic-settings](https://github.com/pydantic/pydantic-settings)?

ci/setup-keycloak.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+set -xe
+
+sudo docker rm -f keycloak
```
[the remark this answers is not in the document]
larsks: @QuanMPhm why the thumbs down? This seems like a valid critique. A shell script with `set -e` active will exit if any command fails with an error.

src/coldfront_plugin_cloud/kc_client.py
```diff
@@ -0,0 +1,77 @@
...
+        r.raise_for_status()
+        r_json = r.json()
```
larsks: Everywhere you are receiving a JSON response you should be using a model:

```
groups = GroupResponse.model_validate(r.json())
return groups[0].id
```

This will make it much more clear what data you're receiving from the API, and will make it easier to access, and it reduces the opportunities for typos.

ci/setup-keycloak.sh
```diff
@@ -0,0 +1,61 @@
...
+
+sudo docker run -d --name keycloak \
+    -e KEYCLOAK_ADMIN=admin \
+    -e KEYCLOAK_ADMIN_PASSWORD=nomoresecret \
+    -p 8080:8080 \
+    -p 8443:8443 \
+    quay.io/keycloak/keycloak:25.0 start-dev
+
+# wait for keycloak to be ready
+until curl -s http://localhost:8080/auth/realms/master; do
```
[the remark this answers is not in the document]
larsks: This seems like a good idea.

ci/setup-keycloak.sh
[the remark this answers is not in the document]
QuanMPhm: @larsks From local testing, running `docker rm` with the `-f` option makes the command return 0 if the container doesn't exist.

When I saw this comment, I remember we also have [a similar statement ](https://github.com/nerc-project/coldfront-plugin-cloud/blob/25a55a2bbcbf06163edb29e77c714516c48176fa/ci/microshift.sh#L14)in one of the other CI files, and quickly tested the command behavior.

src/coldfront_plugin_cloud/kc_client.py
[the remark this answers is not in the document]
QuanMPhm: I see. I'll look at it more closely, now that this PR is not due two days ago.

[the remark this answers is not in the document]
QuanMPhm: @larsks Ok I see your idea now. Initially, my thought was, if we cache the class method, regardless of whether external functions that initialize the class is cached or not, the class function is "permanently" cached, across all instances of the class. Reading some stuff and testing lets me see your point

src/coldfront_plugin_cloud/signals.py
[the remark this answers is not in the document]
QuanMPhm: @knikolla Are you fine adding a pydantic settings module in this PR, or hold this off to a PR that puts ALL the settings in the repo into a settings model?

QuanMPhm:
Addressed all comments, except [one](https://github.com/nerc-project/coldfront-plugin-cloud/pull/249#discussion_r3126172751) waiting for @knikolla response

naved001 (approved):

src/coldfront_plugin_cloud/signals.py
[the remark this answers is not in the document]
knikolla: I'm fine with pushing this into a separate refactoring PR.

knikolla (approved):

PR #261, opened 2025-10-31
Integrate nerc-rates for dynamic outage information loading

### LINKED ISSUE #256: Implement support for pulling outages from `nerc-rates`
https://github.com/CCI-MOC/nerc-rates/pull/42 implements storing outages in `nerc-rates` as YAML files. We should implement support for pulling outage information from there.

jimmysway:
- Add load_outages_from_nerc_rates() to fetch outages from nerc-rates repository
- Update calculate_storage_gb_hours to auto-load outages for "stack" and "ocp-prod" services
- Maintain backwards compatibility with --excluded-time-ranges flag
- Add test coverage for nerc-rates integration

Closes #256

### DIFF
--- a/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
+++ b/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
@@ -20,6 +20,12 @@ logger = logging.getLogger(__name__)
 
 _RATES = None
 
+RESOURCE_NAME_TO_NERC_SERVICE = {
+    "NERC": "stack",
+    "NERC-OCP": "ocp-prod",
+    "NERC-OCP-EDU": "academic",
+}
+
 
 def get_rates():
     # nerc-rates doesn't work with Python 3.9, which is what ColdFront is currently
@@ -142,13 +148,6 @@ class Command(BaseCommand):
             action="store_true",
             help="Upload generated CSV invoice to S3 storage.",
         )
-        parser.add_argument(
-            "--excluded-time-ranges",
-            type=str,
-            default=None,
-            nargs="+",
-            help="List of time ranges excluded from billing, in ISO format.",
-        )
 
     @staticmethod
     def default_start_argument():
@@ -198,8 +197,24 @@ class Command(BaseCommand):
         logger.info(f"Uploaded to {secondary_location}.")
 
     def handle(self, *args, **options):
+        def get_outages_for_service(resource_name: str):
+            """Get outages for a service from nerc-rates.
+
+            :param resource_name: Name of the resource to get outages for.
+            :return: List of excluded intervals or None.
+            """
+            service_name = RESOURCE_NAME_TO_NERC_SERVICE.get(resource_name)
+            if service_name:
+                return utils.load_outages_from_nerc_rates(
+                    options["start"], options["end"], service_name
+                )
+            return None
+
         def process_invoice_row(allocation, attrs, su_name, rate):
             """Calculate the value and write the bill using the writer."""
+            resource_name = allocation.resources.first().name
+            excluded_intervals_list = get_outages_for_service(resource_name)
+
             time = 0
             for attribute in attrs:
                 time += utils.calculate_quota_unit_hours(
@@ -233,13 +248,6 @@ class Command(BaseCommand):
 
         logger.info(f"Processing invoices for {options['invoice_month']}.")
         logger.info(f"Interval {options['start'] - options['end']}.")
-
-        if options["excluded_time_ranges"]:
-            excluded_intervals_list = utils.load_excluded_intervals(
-                options["excluded_time_ranges"]
-            )
-        else:
-            excluded_intervals_list = None
 
         openstack_resources = Resource.objects.filter(
             resource_type=ResourceType.objects.get(name="OpenStack")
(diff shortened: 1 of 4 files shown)


jimmysway:
Based on previous discussions I think the service mapping was something like

OpenStack resources → stack
OpenShift resources → ocp-prod (or should it be both ocp-prod and ocp-test?) 

Is there also other services that need to be included?

knikolla:
Created a related issue as a second step, after this PR is merged, to remove the PR and have the mapping be stored as attributes in the Resources themselves. For now, a dict map in `calculate_storage_hours` is fine. 

https://github.com/nerc-project/coldfront-plugin-cloud/issues/262

knikolla (changes requested):
Review in prior comments. Please implement a dictionary map to fetch the appropriate outage based on cluster name map.

src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -285,15 +291,29 @@ def process_invoice_row(allocation, attrs, su_name, rate):
...
+
             for allocation in openstack_allocations:
                 allocation_str = (
                     f'{allocation.pk} of project "{allocation.project.title}"'
                 )
-                msg = f"Starting billing for allocation {allocation_str}."
-                logger.debug(msg)
+                logger.debug(f"Starting billing for allocation {allocation_str}.")
+
+                if not options["excluded_time_ranges"]:
```
knikolla: This code block is repeated exactly (or almost identically) down below, so I don't think is quite the right place to implement this.

---

On line 243, you have the following code block

```python
if options["excluded_time_ranges"]:
            excluded_intervals_list = utils.load_excluded_intervals(
                options["excluded_time_ranges"]
            )
        else:
            excluded_intervals_list = None
```

This would be a better place for loading the outages. You can split that out from there and define a function as follow.

```python
@functools.cache()
def get_outages_for_service(service):
    if options["excluded_time_ranges"]:
        return utils.load_excluded_intervals(
            options["excluded_time_ranges"]
        )
    else:
        return utils.load_outages_from_nerc_rates(
            options["start"],
            options["end"],
            RESOURCE_NAME_TO_NERC_SERVICE[service_name]
        )
```

You'll need to pass in options or allow the function to fetch them from either other functions or global variables. This will also cache the result. 

---

`process_invoice_row(allocation, attrs, su_name, rate)` on line 211, takes the allocation as a parameter. Therefore you are aware of the allocation, and by accessing `allocation.resources.first().name` you can get the associated Resource name of that allocation, which you can use  to call the get_outages_for_service(service).

---

jimmysway:
> @jimmysway actually, seeing the asymmetry between providing CLI args (apply to all resources) and loaded from `nerc-rates` (apply to the specific resources as listed in the specific outage,) I'm okay with dropping the possibility to provide excluded intervals from the CLI.

@knikolla Are there any downstream consequences for something like this that I should be aware of?

src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
```diff
@@ -87,6 +90,43 @@ def test_new_allocation_quota(self):
                 "2020-03-02 00:00:00,2020-03-03 05:00:00",
             )
 
+    @patch("coldfront_plugin_cloud.utils.load_outages_from_nerc_rates")
+    def test_calculate_storage_with_nerc_rates_outages(self, mock_load_outages):
+        """Test that nerc-rates outages are loaded based on resource name mapping."""
```
QuanMPhm: Building on @knikolla's previous feedback, I would suggest you remove this test, and mock `load_outages_from_nerc_rates` in `test_new_allocation_quota`. You can then make your assertions on `mock_load_outages` there as you see fit

src/coldfront_plugin_cloud/utils.py
```diff
@@ -223,6 +229,24 @@ def check_overlapping_intervals(excluded_intervals_list):
     return excluded_intervals_list
 
 
+def load_outages_from_nerc_rates(
```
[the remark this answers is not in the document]
knikolla: You can just wrap the function in `@functools.cache` and then you don't need to implement the outages_cache you have implemented in the other file.

I would also suggest you move the `load_from_url` call outside of the function so that `load_from_url` isn't called every time the function is called and the output is uncached, but only once per execution of the program.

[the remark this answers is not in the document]
jimmysway: My bad I thought I shouldn't introduce different dependencies.

src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
```diff
@@ -597,3 +606,67 @@ def test_load_excluded_intervals_invalid(self):
...
+        mock_outages = [
+            (
+                pytz.utc.localize(datetime.datetime(2020, 3, 10, 0, 0, 0)),
+                pytz.utc.localize(datetime.datetime(2020, 3, 12, 0, 0, 0)),
+            )
+        ]
+        mock_load_outages.return_value = mock_outages
+        mock_rates_loader.return_value.get_value_at.return_value = Decimal("0.001")
+
+        resource = self.new_openstack_resource(name="NERC")
```
knikolla: Please use a different name and patch the dictionary that does the translation.

knikolla (changes requested):
One final comment, otherwise looks ready to go. Good work.

src/coldfront_plugin_cloud/tests/unit/test_calculate_quota_unit_hours.py
jimmysway: I'm working on an update because I realised I am just mocking the function `load_outages_from_nerc_rates` and we only verified that the wrapper was called successfully. I think I mocking it one level lower is better because then the `load_outages_from_nerc_rates` actually runs  so that the nerc-rates  `get_outages_during` method gets called with correct ISO-formatted dates and service name

knikolla (approved):

knikolla:
@QuanMPhm were your comments resolved?

QuanMPhm (approved):

PR #264, opened 2025-11-05
Validate Openshift limit ranges

### LINKED ISSUE #257: `validate_allocations` should update `LimitRange` if out of sync
`LimitRange` values are applied to an OpenShift namespace only upon creation, so therefore when the default value changes in ColdFront this value is not updated in the existing namespaces.

- `validate_allocations` should update the values that are out of sync.
- This should be implemented in a "Resource" agnostic way

QuanMPhm:
Closes #257. 

A function `set_project_configuration()` has been added to `openshift.py` to validate limit ranges. This allows changes to the default limit ranges (`LIMITRANGE_DEFAULTS`) to be applied to pre-existing Openshift projects.
The plan will be for this function to eventually contain all validation logic for Openshift project configurations, allowing `validation_allocations` to be resource-agnostic.

New functional test case added
`parse_quota_value` has been moved to `utils.py` as it is now used in two places


src/coldfront_plugin_cloud/openshift.py
```diff
@@ -130,6 +130,59 @@ def get_resource_api(self, api_version: str, kind: str):
...
+            logger.info(f"Created default LimitRange for namespace {project_id}.")
+
+        elif len(limits) > 1:
+            logger.warning(
+                f"More than one LimitRange found for namespace {project_id}."
+            )
+            if not dry_run:
+                self._openshift_delete_limits(project_id)
+                self._openshift_create_limits(project_id)
+            logger.info(f"Deleted LimitRanges for namespace {project_id}.")
```
knikolla: Deleted and recreated

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -54,6 +56,94 @@ def clean_openshift_metadata(obj):
...
+            quota_str = round(quota_str / suffix["Gi"])
+    elif quota_str and quota_str == "0":
+        quota_str = 0
+
+    return quota_str
+
+
+def limit_ranges_diff(
+    expected_lr_list: list[dict], actual_lr_list: list[dict], project_id: str = ""
+) -> list[str]:
```
knikolla: You are providing `project_id` solely for the purposes of printing log messages. 

Instead of returning a `list[str]` you can return a `list[LimitRangeDifference]` where `LimitRangeDifference` is a namedtuple with 3 keys (key, expected, actual). This way you don't need to pass in `project_id` and can get back an object that you can reason better about.

For non-default fields you can return `LimitRangeDifference(key="whatever the key", expected=None, actual="actual")`

From the calling function you can check that you got returned an empty list, and if not log the messages by iterating over the list and logging each message.

knikolla (changes requested):

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -54,6 +56,112 @@ def clean_openshift_metadata(obj):
...
+            quota_str = round(quota_str / suffix["Mi"])
+        elif "Storage" in attr:
+            quota_str = round(quota_str / suffix["Gi"])
+    elif quota_str and quota_str == "0":
+        quota_str = 0
+
+    return quota_str
+
+
+class LimitRangeDifference:
```
knikolla: Just use a namedtuple, it gives you `__eq__` for free.

src/coldfront_plugin_cloud/openshift.py
```diff
@@ -54,6 +57,104 @@ def clean_openshift_metadata(obj):
...
+
+        # Extra fields in actual limit range, so else statement should only be expected fields
+        if key not in expected_lr:
+            differences.append(LimitRangeDifference(key, None, actual_lr[key]))
+        else:
+            for resource in expected_lr.setdefault(key, {}) | actual_lr.setdefault(
+                key, {}
+            ):
+                # Extra nested field in actual lr
+                if resource not in expected_lr[key]:
```
knikolla: Why is this code block required? In which instances would the comparison in the `else` block not work?

knikolla (changes requested):

knikolla (changes requested):

knikolla (approved):

Issue #258, opened 2025-10-22
Further lower minimum limitranges or remove them


naved001:
Right now minimum CPU is 25m and right now we are running into an issue with cert-manager trying to run a pod with a 10m minimum. 

We could: 

1. lower the minimum to absolute lowest - CPU to 1m and memory to 1 byte, so users cannot just set it to 0. 
2. Remove the minimum entirely and in the billing code just set a minimum we will always bill for.

PR #268, opened 2025-11-19
Further lower the minimum required CPU and memory

naved001:
This sets it to 1 milicore CPU and 1 Kibibyte of memory. I don't think anything can run with just 1 KiB of memory so I think this is quite low at this point.
 
 Is there a reason why this is part of the code and not part of some config file?

closes https://github.com/nerc-project/coldfront-plugin-cloud/issues/258

### DIFF
--- a/src/coldfront_plugin_cloud/openshift.py
+++ b/src/coldfront_plugin_cloud/openshift.py
@@ -40,7 +40,7 @@ LIMITRANGE_DEFAULTS = [
         "type": "Container",
         "default": {"cpu": "1", "memory": "4096Mi", "nvidia.com/gpu": "0"},
         "defaultRequest": {"cpu": "500m", "memory": "2048Mi", "nvidia.com/gpu": "0"},
-        "min": {"cpu": "25m", "memory": "32Mi"},
+        "min": {"cpu": "1m", "memory": "1Ki"},
     }
 ]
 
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/openshift.py
```diff
@@ -40,7 +40,7 @@
         "type": "Container",
         "default": {"cpu": "1", "memory": "4096Mi", "nvidia.com/gpu": "0"},
         "defaultRequest": {"cpu": "500m", "memory": "2048Mi", "nvidia.com/gpu": "0"},
-        "min": {"cpu": "25m", "memory": "32Mi"},
+        "min": {"cpu": "1m", "memory": "1Ki"},
```
QuanMPhm: The CI is failing because we have a unit test that tests a validation function for OpenShift limits, [which tests exactly what the minimum for memory should be](https://github.com/nerc-project/coldfront-plugin-cloud/blob/df2cc66f7b6523b6a4fb343570e4b9bdcb7badc5/src/coldfront_plugin_cloud/tests/unit/test_openshift_utils.py#L63-L65)

QuanMPhm (changes requested):

src/coldfront_plugin_cloud/openshift.py
naved001: Thanks for pointing it out. I've fixed it.

PR #277, opened 2025-12-02
Add attributes to tracking allocation usage

### LINKED ISSUE #270: Add private billing-related allocation attributes
As described in the daily usage showback document, we want to add several billing-related attributes to help keep track of each allocation's monetary costs and alert users:
- `Daily Charges for Month`: A private JSON-encoded string containing a list of key-value pairs that shows the daily costs for an allocation by SU type:
```
"daily": {
	"2025-10-02": {
		"OpenShift CPU": "3.60",
		"OpenShift GPUA100SXM4": "124.00"
		},
	"2025-10-03": {
		"OpenShift CPU": "10.65",
		"OpenShift GPUA100SXM4": "326.00"
		}
	}
}

```
- `Project Alert` & `Allocation Alert `: A dollar amount at which the PI would want to be alerted. If either attributes are present, the PI should be alerted once all-time costs reaches said amount
- `Previous Charges` - Will store in JSON format a key-value pair of the charges for previous months, as defined in the final invoices.
```
"previous": {
	"2025-09": {
		"OpenShift CPU": "10.65",
		"OpenShift GPUA100SXM4": "326.00"
	}
}
```

QuanMPhm:
Closes #270. I've introduced some pydantic models to load and validate the usage dicts. These attributes are not yet used in any way in any part of the code.

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -54,6 +54,9 @@ RESOURCE_ATTRIBUTES = [
 ALLOCATION_PROJECT_ID = "Allocated Project ID"
 ALLOCATION_PROJECT_NAME = "Allocated Project Name"
 ALLOCATION_INSTITUTION_SPECIFIC_CODE = "Institution-Specific Code"
+ALLOCATION_CUMULATIVE_CHARGES = "Cumulative Daily Charges for Month"
+ALLOCATION_PREVIOUS_CHARGES = "Previous Charges"
+ALLOCATION_ALERT = "Monthly Allocation Cost Alert"
 
 ALLOCATION_ATTRIBUTES = [
     CloudAllocationAttribute(
@@ -65,6 +68,19 @@ ALLOCATION_ATTRIBUTES = [
     CloudAllocationAttribute(
         name=ALLOCATION_INSTITUTION_SPECIFIC_CODE, type="Text", is_changeable=True
     ),
+    CloudAllocationAttribute(
+        name=ALLOCATION_CUMULATIVE_CHARGES,
+        type="Text",
+        is_private=True,
+        is_changeable=True,
+    ),
+    CloudAllocationAttribute(
+        name=ALLOCATION_PREVIOUS_CHARGES,
+        type="Text",
+        is_private=True,
+        is_changeable=True,
+    ),
+    CloudAllocationAttribute(name=ALLOCATION_ALERT, type="Int", is_changeable=True),
 ]
 
 ###########################################################
--- a/src/coldfront_plugin_cloud/tests/unit/test_usage_models.py
+++ b/src/coldfront_plugin_cloud/tests/unit/test_usage_models.py
@@ -1,0 +1,65 @@
+from decimal import Decimal
+from pydantic import ValidationError
+
+from coldfront_plugin_cloud import usage_models
+from coldfront_plugin_cloud.tests import base
+
+
+class TestUsageModels(base.TestBase):
+    def test_usage_info(self):
+        # valid: values coerced to Decimal
+        ui = usage_models.UsageInfo(
+            root={"su-a": Decimal("1.5"), "su-b": 2, "su-c": "3.25"}
+        )
+        self.assertIsInstance(ui.root, dict)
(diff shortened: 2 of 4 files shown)


knikolla (changes requested):

src/coldfront_plugin_cloud/tests/unit/test_usage_models.py
```diff
@@ -0,0 +1,50 @@
...
+from pydantic import ValidationError
+
+from coldfront_plugin_cloud import usage_models
+from coldfront_plugin_cloud.tests import base
+
+
+class TestUsageModels(base.TestBase):
+    def test_usage_info(self):
+        # valid: values coerced to Decimal
+        ui = usage_models.UsageInfo(root={"su-a": 1.5, "su-b": 2, "su-c": "3.25"})
```
knikolla: Why not right this directly as Decimals rather than converting them from floats?

src/coldfront_plugin_cloud/tests/unit/test_usage_models.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEqual(ui.root["su-b"], 2)
+        self.assertEqual(ui.root["su-c"], 3.25)
+
+        # invalid: non-numeric string should raise ValidationError
+        with self.assertRaises(ValidationError):
+            usage_models.UsageInfo(root={"su-x": "not-a-number"})
+
+    def test_daily_charges_dict(self):
+        # Valid DailyChargesDict with YYYY-MM-DD keys
+        data = {
```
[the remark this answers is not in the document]
knikolla: Please update this to only include days from a single month as would be the valid data.

knikolla (changes requested):

knikolla (approved):

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -54,6 +54,9 @@ class CloudAllocationAttribute:
 ALLOCATION_PROJECT_ID = "Allocated Project ID"
 ALLOCATION_PROJECT_NAME = "Allocated Project Name"
 ALLOCATION_INSTITUTION_SPECIFIC_CODE = "Institution-Specific Code"
+ALLOCATION_CUMULATIVE_CHARGES = "Cumulative Daily Charges for Month"
+ALLOCATION_PREVIOUS_CHARGES = "Previous Charges"
+ALLOCATION_ALERT = "Monthly Allocation Cost Alert"
```
knikolla: @joachimweyl can you please sign off on the naming of these attributes?

src/coldfront_plugin_cloud/usage_models.py
```diff
@@ -0,0 +1,70 @@
...
+        """Leverage lexical ordering of YYYY-MM-DD and YYYY-MM strings."""
+        return max(self.root.keys()) if self.root else ""
+
+
+class CumulativeChargesDict(ChargesDict[DateField]):
+    @pydantic.model_validator(mode="after")
+    def check_month(self):
+        # Ensure all keys are in the same month
+        if self.root:
+            months = {date_str[:7] for date_str in self.root.keys()}
```
naved001: This will probably be not a problem as all the input data will be in YYYY-MM-DD or YYYY-MM, but since you are strictly validting types and such I though I should mention this:


You are validating dates in `validate_date_str` with 

```
datetime.datetime.strptime(v, "%Y-%m-%d")
```

And according to that both, `2025-05-01` and `2025-5-1`, are valid dates. However, here you are relying on string splicing to get `2025-05`, but it'll return `2025-5-` for the other one. 

Personally, I'd get a datetime object and do a `.month` on it.

```
>>> datetime.strptime("2025-5-1", "%Y-%m-%d").month
5
>>> datetime.strptime("2025-05-01", "%Y-%m-%d").month
5
```

I guess I am just not a fan of using string splicing after I've already validated that it's a date.

naved001 (approved):
Seems fine, just one nit picky comment.

src/coldfront_plugin_cloud/attributes.py
knikolla: Those are just the names of the variables and we can change those without user facing impact at any point.

PR #286, opened 2026-01-13
Use postgres for CI run_unit_tests

### LINKED ISSUE #254: Change to using Postgres database for testing
During testing, it was found that Coldfront's default MySQL database does not enforce a [character limit set by Coldfront](https://github.com/ubccr/coldfront/blob/627e9ea957262b3bfcb37e6a19da3d8271a497f1/coldfront/core/allocation/models.py#L425). Our production Coldfront uses Postgres, which does enforce the limit. This caused an error while trying to ad the new IBM Storage attribute. More details in this [PR](https://github.com/nerc-project/coldfront-plugin-cloud/pull/253).

To better mimic the production environment, our tests should use a Postgres database instead.

QuanMPhm:
Closes #254.

A Postgres-specific bug was found regarding creating Field Of Science entries. More details in `tests/unit/test_migrate_field_of_science.py`.

### DIFF
--- a/.github/workflows/test-functional-microstack.yaml
+++ b/.github/workflows/test-functional-microstack.yaml
@@ -21,7 +21,7 @@ jobs:
         with:
           python-version: 3.12
 
-      - name: Install dependancies, ColdFront and plugin
+      - name: Install dependencies, ColdFront and plugin
         run: |
           ./ci/setup.sh
 


QuanMPhm:
@knikolla I've made the one test case in `src/coldfront_plugin_cloud/tests/unit/test_migrate_field_of_science.py` check if `DB_URL` is set, in which case it will run the postgres specific SQL commands. Otherwise, all the other tests can still be ran with SQLite, you just need to not set `DB_URL`

PR #317, opened 2026-05-26
Add AllocationDailyBillableUsage model and improve tests

### LINKED ISSUE #292: Create new table `AllocationDailyBillableUsage`
- Create new table `AllocationDailyBillableUsage` and corresponding migration.
- Extend the `fetch_daily_billable_usage.py` command to populate the table.
- Extend the `fetch_daily_billable_usage.py` with an extra argument `--remove` that removes the rows for a specific date from the table.

jimmysway:
Closes #292


src/coldfront_plugin_cloud/apps.py
```diff
@@ -6,3 +6,4 @@ class OpenStackConfig(AppConfig):
 
     def ready(self):
         import coldfront_plugin_cloud.signals  # noqa: F401
+        import coldfront_plugin_cloud.models.daily_billable_usage  # noqa: F401
```
knikolla: why is this being imported here?

jimmysway: I don't really understand everything here but I stumbled across this article when trying to figure out why .signals was being imported which led me to this article in the django documentation 

https://docs.djangoproject.com/en/6.0/ref/applications/#how-applications-are-loaded

Where it says 

> Then Django attempts to import the models submodule of each application, if there is one.
> 
> You must define or import all models in your application’s models.py or models/__init__.py. Otherwise, the application registry may not be fully populated at this point, which could cause the ORM to malfunction.
> 
> Once this stage completes, APIs that operate on models such as [get_model()](https://docs.djangoproject.com/en/6.0/ref/applications/#django.apps.apps.get_model) become usable.

We have historically left the `__init__.py` blank and I don't know if this `daily_billable_usage` model is fundamentally different than the other ones because it is `ORM` related? I might need some help understanding this

src/coldfront_plugin_cloud/management/commands/seed_daily_billable_usage.py
```diff
@@ -0,0 +1,153 @@
...
+        ramped[su_type] = str(float(amount) + day_index * step)
+    return usage_models.UsageInfo(ramped)
+
+
+def current_month() -> str:
+    today = date.today()
+    return f"{today.year}-{today.month:02d}"
+
+
+class Command(BaseCommand):
```
knikolla: Please don't create a new command for this. Instead look into either mocking the tests so that the values that you need are returned from the fictional CSV or sample values inserted into the DB.

src/coldfront_plugin_cloud/apps.py
knikolla: Would adding the import in `models/__init__.py` instead of here work?

knikolla (changes requested):
Did a first pass at reviewing timeboxed to 25mins. Overall looks in the right direction. Please instead of creating a new command enhance the testing code to create the testing setup you need.

src/coldfront_plugin_cloud/management/commands/seed_daily_billable_usage.py
jimmysway: Just to clarify, this seeding logic is primarily for testing of the downstream graphical UI elements in `coldfront-nerc`. I will find another way to insert sample values directly into the DB

jimmysway:
I will go over all the comments but I also have one question right now there is a `management/commands/fetch_daily_billable_usage.py` which is the fetching code for S3 and also my newly written `models/daily_billable_usage.py` which is fetching from the database.

I find it confusing sometimes for which is which do you think I should rename these to distinguish between them better or is it fine.

jimmysway:
I will also update the table implementation to use this injection pattern instead

scripts/seed_sample_daily_billable_usage_shell.py
QuanMPhm, line 1: If this file is intended to inject test data that will be used to test the UI graphs in `coldfront-nerc`, shouldn't this file be added in that PR instead? I'm not familiar with the UI code or how testing will be done with it yet, but that's my reaction. It's odd seeing a file added in a PR that doesn't use it.

@jimmysway Is this only way to test the UI graphs?

src/coldfront_plugin_cloud/models/daily_billable_usage.py
```diff
@@ -0,0 +1,36 @@
...
+from coldfront.core.allocation.models import Allocation
+
+
+class AllocationDailyBillableUsage(TimeStampedModel):
+    """Stores daily billable usage for allocations by SU type."""
+
+    allocation = models.ForeignKey(
+        Allocation,
+        on_delete=models.CASCADE,
+        related_name="daily_usage_records",
```
QuanMPhm, line 12: Why do we want `related_name` to be `daily_usage_records`?

src/coldfront_plugin_cloud/models/daily_billable_usage.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    class Meta:
+        db_table = "coldfront_plugin_cloud_allocationdailybillableusage"
+        unique_together = [["allocation", "date", "su_type"]]
+        indexes = [
+            models.Index(fields=["allocation", "date"]),
+            models.Index(fields=["date"]),
+        ]
+        ordering = ["-date", "allocation", "su_type"]
```
QuanMPhm, line 33: aside from `unique_together`, why are any of the other `Meta` options nessecary? I suppose `indexes` gives a performance boost.

`unique_together` can also be simplified to `["allocation", "date", "su_type"]`, no need for nested list

src/coldfront_plugin_cloud/billable_usage.py
```diff
@@ -0,0 +1,145 @@
...
+        ...     AllocationDailyBillableUsage(su_type="Storage", value=30.12),
+        ... ]
+        >>> info = _rows_to_usage_info(rows)
+        >>> info.root["OpenStack CPU"]
+        Decimal('100')
+        >>> info.total_charges
+        Decimal('130.12')
+    """
+    if rows is None:
+        raise TypeError("rows must not be None")
```
QuanMPhm: I impression with the majority of the raised errors in this module is that they are unnecessary. This one, for example, wouldn't be anymore helpful than the `TypeError` we'll get below from `for row in rows:`

I would only raise an error if the code encounters a case where we have deliberately decided that the expected behavior should be a raised error. I think raising errors like in this example only adds a slightly more helpful error message, and more mental burden when maintaining and reviewing code. @knikolla Do you have an opinion?

src/coldfront_plugin_cloud/billable_usage.py
```diff
@@ -0,0 +1,145 @@
...
+
+    usage_info = UsageInfo({})
+    for row in rows:
+        if not isinstance(row, AllocationDailyBillableUsage):
+            raise TypeError(
+                f"each row must be AllocationDailyBillableUsage, got {type(row).__name__}"
+            )
+        if not row.su_type:
+            raise ValueError(f"usage row id={row.pk} has empty su_type")
```
QuanMPhm, line 38: When would `su_type` be empty, and do we want to have empty `su_types` (`""`)

src/coldfront_plugin_cloud/billable_usage.py
```diff
@@ -0,0 +1,145 @@
...
+        Decimal('180.12')
+    """
+    if not isinstance(allocation, Allocation):
+        raise TypeError(
+            f"allocation must be Allocation, got {type(allocation).__name__}"
+        )
+    if allocation.pk is None:
+        raise ValueError("allocation must be saved (have a primary key)")
+    if not isinstance(date, str) or not date.strip():
+        raise ValueError("date must be a non-empty YYYY-MM-DD string")
```
QuanMPhm: I think this is unnecessary since `validate_date_str` would validate the string anyways

src/coldfront_plugin_cloud/billable_usage.py
```diff
@@ -0,0 +1,145 @@
...
+            raise TypeError(
+                f"each row must be AllocationDailyBillableUsage, got {type(row).__name__}"
+            )
+        if not row.su_type:
+            raise ValueError(f"usage row id={row.pk} has empty su_type")
+        usage_info.root[row.su_type] = row.value
+    return usage_info
+
+
+def get_daily_billable_usage(allocation: Allocation, date: str) -> UsageInfo:
```
QuanMPhm: Can't we remove this function, since `get_daily_billable_usage_by_date` can accomplish the same thing by passing the same date to both `start_date` and `end_date`?

Alternatively, why not use this in `get_daily_billable_usage_by_date`? You make a few more queries, but it usually won't be more than a month of queries, yes?

.gitignore
```diff
@@ -134,3 +134,4 @@ dmypy.json
 [._]s[a-rt-v][a-z]
 [._]ss[a-gi-z]
 [._]sw[a-p]
+*.db
```
QuanMPhm, line 137: I don't think this should be here. I don't think we'll ever need to commit a `.db` file, but this doesn't seem nessecary

src/coldfront_plugin_cloud/tests/unit/test_fetch_daily_billable_usage.py
QuanMPhm, line 1: @knikolla It seems if we were to run `fetch_daily_billable_usage` twice on the same day, and the S3 invoicing data has changed such that a project is gone from the invoices, then the project's usage entry would still persist in the Coldfront database, unless we run `--remove` and rerun  `fetch_daily_billable_usage` normally again. This is fine?

src/coldfront_plugin_cloud/tests/unit/test_daily_billable_usage.py
```diff
@@ -0,0 +1,329 @@
...
+    def test_multiple_rows(self):
+        # Verify multiple ORM rows collapse into a single SU->Decimal mapping with a correct total.
+        rows = [
+            AllocationDailyBillableUsage(su_type="OpenStack CPU", value=Decimal("100")),
+            AllocationDailyBillableUsage(su_type="Storage", value=Decimal("30.12")),
+        ]
+        usage = _rows_to_usage_info(rows)
+        self.assertEqual(usage.root["OpenStack CPU"], Decimal("100"))
+        self.assertEqual(usage.root["Storage"], Decimal("30.12"))
+        self.assertEqual(usage.total_charges, Decimal("130.12"))
```
QuanMPhm: For all of these test cases, I think its more clear to just create a `UsageInfo` or `dict` "answer", and compare them directly

QuanMPhm (changes requested):
The structure looks good to me. I have comments on some lower-level details

scripts/seed_sample_daily_billable_usage_shell.py
jimmysway: Well this file tests the injection pattern that the UI charts will get their data from, injecting sample data into a database tests the entire flow rather than just mocking it at a higher level. 

I can for sure add this to another PR but it is very dependent on the contents of this PR so it was the easiest way for me to write a script to write some sample data into the db

src/coldfront_plugin_cloud/models/daily_billable_usage.py
jimmysway: It lets you write `allocation.daily_usage_records` to get all the usage rows for a given allocation, instead of Django's ugly default name (`allocation.allocationdailybillableusage_set`). It's just for readability/convenience when querying.

src/coldfront_plugin_cloud/models/daily_billable_usage.py
jimmysway: `db_table = "coldfront_plugin_cloud_allocationdailybillableusage"`

Will be auto generated by django yes, but the preference for being explicit to navigate the db without a lot of prior understanding behind the magic autogeneration pattern that django would do on its own

src/coldfront_plugin_cloud/billable_usage.py
jimmysway: It's slightly different because the return types are different so the single day call is `get_daily_billable_usage_by_date(allocation, date, date).get(date, UsageInfo({}))`. But I think that's fine I will make the change

src/coldfront_plugin_cloud/billable_usage.py
jimmysway: I think we should drop the `isinstance(allocation, Allocation)` and the `if rows is None and isinstance(row, ...)` because these will get defaultly picked up by Python

PR #318, opened 2026-05-28
Resolve breaking changes by kubernetes package

QuanMPhm:
First spotted by @jimmysway in #317 

Release v36.0.0 of the Kubernetes python client package introduced a breaking change to how auth is configured [1]

Auth is now configured using `BearerToken` instead of `authorization`

[1] https://github.com/kubernetes-client/python/blob/release-36.0/CHANGELOG.md#breaking-change-from-upgrading-openapi-generator-to-v660


knikolla (changes requested):
Need to also set the minimum package version to be the one that introduced the breaking change

knikolla (approved):

PR #319, opened 2026-06-04
Added Boolean fied to indicate if project is externally funded

### LINKED ISSUE #320: Capture whether funding is internal or external for a specific coldfront project via a boolean attribute
We should add a boolean field at the Coldfront project level to mark whether a Coldfront project is funded by external sources.

QuanMPhm:
Closes #320. As part of allowing NERC projects to indicate if they are externally funded.

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -20,6 +20,18 @@ class CloudAllocationAttribute:
     is_changeable: bool = True
 
 
+@dataclass
+class CloudProjectAttribute:
+    """Class for configuring Cloud Project Attributes"""
+
+    name: str
+    type: str = "Int"
+    has_usage: bool = False
+    is_private: bool = False
+    is_changeable: bool = True
+    is_unique: bool = True
+
+
 RESOURCE_AUTH_URL = "Identity Endpoint URL"
 RESOURCE_API_URL = "OpenShift API Endpoint URL"
 RESOURCE_IDENTITY_NAME = "OpenShift Identity Provider Name"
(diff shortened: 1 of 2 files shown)


src/coldfront_plugin_cloud/attributes.py
```diff
@@ -120,3 +132,12 @@ class CloudAllocationAttribute:
 ALLOCATION_QUOTA_ATTRIBUTES = [
     CloudAllocationAttribute(name=QUOTA_GPU),
 ]
+
+
+PROJECT_IS_EXTERNALLY_FUNDED = "Is Externally Funded"
+
+PROJECT_ATTRIBUTES = [
+    CloudProjectAttribute(
+        name=PROJECT_IS_EXTERNALLY_FUNDED, type="Yes/No", is_changeable=True
```
knikolla: I don't think this should be is_changeable as it shouldn't be edited by PIs, but set by admins.

Also, I don't think we expose Project Attributes via the API currently, no?

So there would be no way for invoicing to consume this.

knikolla (changes requested):

src/coldfront_plugin_cloud/attributes.py
QuanMPhm: Given your [comment in this PR](https://github.com/nerc-project/coldfront-plugin-api/pull/69#discussion_r3382964535), I assume your question here has been answered? Unless I am missing something else?

QuanMPhm: I have set `is_changeable` to False

jimmysway (approved):

knikolla (approved):

PR #324, opened 2026-08-05
Move hardcoded quota names from attributes.py to tests

### LINKED ISSUE #298: Refactor the `attributes.py` to remove hardcoded quota names
As [mentioned](https://github.com/nerc-project/coldfront-plugin-cloud/pull/287#discussion_r2800426708) by @knikolla during development of the dynamic quota feature, the hardcoded quota attributes in the `attributes.py` file should be moved to tests to indicate that these strings should ever be used in tests.

I will save this issue for the new intern

marcoagonzales007:
Fixes #298 (Migrated to [CCI-MOC/MOC-issues#256](https://github.com/CCI-MOC/MOC-issues/issues/256)) )

This moves the Quota_* constants into a test only module( `tests/attributes.py`) and refactors the remaining uses 

- `openstack.py` finds the Swift object quota by its object instead of by display name.

- `calculate_storage_gb_hours` bills OpenShift storage from each resource's quota specs grouped by invoice_name, like the OpenStack loop already does. 

- `register_default_quotas` inlines the default display names it registers.

- `convert_swift_quota_to_gib` also had a broken reference to the removed `openstack.QUOTA_KEY_MAPPING` 

- `QUOTA_GPU `stays in attributes.py because it's still a statically registered allocation attribute used by count_gpu_usage.

### DIFF
--- a/src/coldfront_plugin_cloud/attributes.py
+++ b/src/coldfront_plugin_cloud/attributes.py
@@ -87,34 +87,8 @@ ALLOCATION_ATTRIBUTES = [
     CloudAllocationAttribute(name=ALLOCATION_ALERT, type="Int", is_changeable=True),
 ]
 
-###########################################################
-# OpenStack Quota Attributes
-QUOTA_INSTANCES = "OpenStack Compute Instance Quota"
-QUOTA_RAM = "OpenStack Compute RAM Quota (MiB)"
-QUOTA_VCPU = "OpenStack Compute vCPU Quota"
-
-QUOTA_VOLUMES = "OpenStack Number of Volumes Quota"
-QUOTA_VOLUMES_GB = "OpenStack Volume Quota (GiB)"
-
-QUOTA_FLOATING_IPS = "OpenStack Floating IP Quota"
-QUOTA_NETWORKS = "Openstack Network Quota"
-
-QUOTA_OBJECT_GB = "OpenStack Swift Quota (GiB)"
-
+# QUOTE_GPU is still registered because count_gpu_usage command relies on it
 QUOTA_GPU = "OpenStack GPU Quota"
-
-###########################################################
-# OpenShift Quota Attributes
-QUOTA_LIMITS_CPU = "OpenShift Limit on CPU Quota"
-QUOTA_LIMITS_MEMORY = "OpenShift Limit on RAM Quota (MiB)"
-QUOTA_LIMITS_EPHEMERAL_STORAGE_GB = "OpenShift Limit on Ephemeral Storage Quota (GiB)"
-QUOTA_REQUESTS_NESE_STORAGE = "OpenShift Request on NESE Storage Quota (GiB)"
-QUOTA_REQUESTS_IBM_STORAGE = "OpenShift Request on IBM Storage Quota (GiB)"
-QUOTA_REQUESTS_GPU = "OpenShift Request on GPU Quota"
-QUOTA_REQUESTS_VM_GPU_A100_SXM4 = "OpenShift Request on GPU A100 SXM4"
-QUOTA_REQUESTS_VM_GPU_V100 = "OpenShift Request on GPU V100"
-QUOTA_REQUESTS_VM_GPU_H100 = "OpenShift Request on GPU H100"
-QUOTA_PVC = "OpenShift Persistent Volume Claims Quota"
 
 
 ALLOCATION_QUOTA_ATTRIBUTES = [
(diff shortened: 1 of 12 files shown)


src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
```diff
@@ -322,7 +322,6 @@ def process_invoice_row(allocation, attrs, su_name, rate):
                 f, delimiter=",", quotechar="|", quoting=csv.QUOTE_MINIMAL
             )
             csv_invoice_writer.writerow(InvoiceRow.get_headers())
-
```
QuanMPhm, line 325: Stray whitespace removal

src/coldfront_plugin_cloud/attributes.py
```diff
@@ -87,35 +87,9 @@ class CloudAllocationAttribute:
...
-QUOTA_VOLUMES = "OpenStack Number of Volumes Quota"
-QUOTA_VOLUMES_GB = "OpenStack Volume Quota (GiB)"
-
-QUOTA_FLOATING_IPS = "OpenStack Floating IP Quota"
-QUOTA_NETWORKS = "Openstack Network Quota"
-
-QUOTA_OBJECT_GB = "OpenStack Swift Quota (GiB)"
-
+# QUOTE_GPU is still registered because count_gpu_usage command relies on it
 QUOTA_GPU = "OpenStack GPU Quota"
```
QuanMPhm, line 91: Thank you for the comment! It helps clarify why we still have one attribute

src/coldfront_plugin_cloud/openstack.py
```diff
@@ -222,12 +222,12 @@ def _set_object_quota(self, project_id, payload):
...
             # converting to bytes for Swift.
-            _, obj_q_mapping = self._extract_quota_label(
-                self.resource_quotaspecs.root[attributes.QUOTA_OBJECT_GB]
-            )
-            payload[obj_q_mapping] *= GB_IN_BYTES
-            if payload[obj_q_mapping] <= 0:
-                payload[obj_q_mapping] = 1
+            for obj_q_mapping in self._get_resource_quota_labels_by_service("object"):
+                if obj_q_mapping not in payload:
+                    continue
```
QuanMPhm, line 227: As it stands, this function expects the object service to only have one quota (`object.x-account-meta-quota-bytes`), as such, it looks misleading to use a for loop. This function would likely cause unintended behavior if there was ever more than one object quota. I would suggest:
```python
obj_q_mapping = self._get_resource_quota_labels_by_service("object")[0]
```

src/coldfront_plugin_cloud/openstack.py
```diff
@@ -291,10 +291,7 @@ def get_quota(self, project_id):
 
         quotas = self._get_network_quota(quotas, project_id)
 
-        if object_quotaspec := self.resource_quotaspecs.root.get(
-            attributes.QUOTA_OBJECT_GB
-        ):
-            _, key = self._extract_quota_label(object_quotaspec)
+        for key in self._get_resource_quota_labels_by_service("object"):
```
QuanMPhm, line 294: This does look aesthetically neater, but could lead to the same confusion mentioned above.

src/coldfront_plugin_cloud/openstack.py
QuanMPhm: Also, why was there a need for the `if` clause?

src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py
QuanMPhm, line 1: Removing the hardcoded quota strings from this file will be more complicated, as we are trying to do a more significant refactor. Refer to this [comment](https://github.com/nerc-project/coldfront-plugin-cloud/pull/287#discussion_r2817705459) to understand the bigger idea of what we are trying to achieve. 

In summary, we want to encode in the quota display name information about the key of the quota on `nerc-rates`. This will require making a coordinated change with `nerc-rates`. We eventually want this command to be in such a state that there should be no code that is statically fetching certain storage rates (ibm-storage), but is dynamically fetching when only given the storage quotaspec in each allocation. 

This refactor will be somewhat significant, so for now, I would suggest that you revert this file, and replace references to `attributes` with just hardcoded string literals or top-level constants. You can refactor this file in another PR.

QuanMPhm:
@marcoagonzales007 Please investigate the CI failures

